PR #48, opened 2022-03-23
DDF feedback fixes

PELN:
This PR includes some changes/fixes for some of the components in dpl-design-system
- Removed props from Modal Search since they were not used for anything
- Add possibility for changing the text in the "Text Modal" in storybook
- Fix position error with link-filters in Material Banner/Modal Details
- Add option to choose icons on dropdown, since it should be able to change between two icons
- Create new component "Modal-pause" since it was missing
- Disable controls in checkbox and materialCard, since we can't get the states to work with controls in SB. The only help to find on the internet is to make states work with "Actions" - not "Controls".
- Add datepicker input to ModalPause component

The error with Chromatic is not from this PR. I have built the storybook locally and reviewed the changes there.

### DIFF
--- a/src/stories/modal-pause/modal-pause.scss
+++ b/src/stories/modal-pause/modal-pause.scss
@@ -1,0 +1,75 @@
+$modal-pause-container-width: 550px;
+
+.modal-pause {
+  flex-direction: column;
+  align-items: center;
+}
+
+.modal-pause__container {
+  margin: 0 24px;
+  padding-top: 96px;
+  max-width: $modal-pause-container-width;
+  width: 100%;
+
+  @include breakpoint-s {
+    margin: 0 45px;
+  }
+}
+
+.modal-pause__button {
+  display: grid;
+  justify-content: center;
+}
+
+.modal_pause__text-link {
+  .link-tag {
+    font-size: 14px;
+    padding-left: 5px;
+  }
+}
+
+.datepickers {
+  display: flex;
+  justify-content: space-between;
+  margin-bottom: 80px;
+
+  input {
+    border: 1px solid $c-global-tertiary-1;
+    box-sizing: border-box;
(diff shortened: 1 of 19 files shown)


src/stories/modal-pause/modal-pause.scss
```diff
@@ -0,0 +1,75 @@
+$MODAL_PAUSE_CONTAINER_WIDTH: 550px;
```
Adamik10, line 1: Mmm. We are already not consistent with this, because we use caps some places and not others, but maybe for a new component I would just make it not capitalized.

src/stories/modal-pause/modal-pause.scss
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  .datepicker {
+    width: 100%;
+    display: flex;
+    flex-direction: column;
+  }
+
+  [type="date"] {
+    background: #f6f5f0 url("https://cdn1.iconfinder.com/data/icons/cc_mono_icon_set/blacks/16x16/calendar_2.png") 90% 50% no-repeat ;
```
Adamik10, line 58: Do we have a variable for this color somewhere?

Adamik10 (changes requested):
It looks good :) I have a couple of comments. 
In regards to Dropdown.tsx, I think it may be a bit busy with a component inside the component? But I can't think of how else I would do it, so I didn't leave any comments there.

src/stories/modal-pause/modal-pause.scss
PELN: I have fixed it 🙌

src/stories/modal-pause/modal-pause.scss
PELN: Good catch, there was a variable that I could use 👏

PR #52, opened 2022-05-18
Merge Feature/header into setup branch

Adamik10:
This feature branch updates the header component to comply with the newest figma mockups. The component already existed beforehand - thus changes are minimal.

Ticket: [DDFSOEG-96](https://reload.atlassian.net/browse/DDFSOEG-96)


cableman (changes requested):
Why is the package-lock file update when the actual change is scss only?

src/stories/header/header.scss
```diff
@@ -149,6 +149,12 @@
   ::placeholder {
     color: $c-text-secondary-gray;
   }
+
```
sinejespersen, line 152: I am not sure how this looks "in the wild", but be aware that if it removes the possibility to see if an element is in focus, then it does not comply with WCAG2.1 https://www.w3.org/WAI/WCAG21/Understanding/focus-visible.html

Adamik10: yes @sinejespersen. In the mockups, focus state is shown by changing the background color of the input field to white without it being outlined. So the whole field changes color from:
image
to:
image

Adamik10:
> Why is the package-lock file update when the actual change is scss only?

@cableman It is because I didn't commit these things with previous changes, where we introduced clsx package and removed engines. And THAT's because I thought package-lock.json should have been a part of git ignore, which now I know it's ofc indeed not. I'm also fairly new to git, so it is entirely possible that these changes already lay somewhere on another branch, and I should have just rebased them into here.

src/stories/header/header.scss
sinejespersen: Sure is, thanks.

PR #54, opened 2022-05-27
Autosuggest-text OLD

Adamik10:
### **Description**

The **autosuggest text component** is a part of the autosuggest dropdown as one of three separate parts:
- autosuggest text
- autosuggest material
- autosuggest topic

This component's styling is closely bound to the search bar and header, which it depends on for styling. Thus I include a dummy search bar and header in the story. This way the _autosuggest text_ component doesn't contain anything extra/unnecessary apart from itself, but the story shows it in the context with proper styling. 

### **Screenshots**
image
image

### **Comments/Questions**


src/stories/autosuggest-text/autosuggest-text.scss
```diff
@@ -0,0 +1,28 @@
+.suggest-text {
```
sinejespersen, line 1: I think, as I have understood BEM, container ought to be an element and not a modifier? and then nest the item inside the container...

sinejespersen: Perhaps also rename to "autosuggest-text" so it fits the react component? I think this is difficult though, to have names that fit components, as there will be another auto-suggest-component in the dpl-react repo which would also have to have the same name if we would like to enforce consistency in naming?

src/stories/autosuggest-text/AutosuggestText.stories.tsx
```diff
@@ -0,0 +1,54 @@
...
+    items: {
+      name: "Items",
+      defaultValue: ["Item one", "Item two"],
+      control: { type: "array" }
+    }
+  },
+  parameters: {},
+} as ComponentMeta<typeof AutosuggestText>;
+
+const Template: ComponentStory<typeof AutosuggestText> = (args) => (
```
spaceo, line 19: To be consistent I think you sgould import the AutosuggestTextProps type and attach it to args: 
`(args: AutosuggestTextProps)`
Like it is done here:
https://github.com/danskernesdigitalebibliotek/dpl-design-system/blob/main/src/stories/breadcrumb/Breadcrumb.stories.tsx#L28

src/stories/header/Header.tsx
```diff
@@ -90,16 +90,18 @@ export const Header = (props: HeaderProps) => {
...
-              <img
-                className="header__menu-search-icon"
-                src={`icons/basic/icon-search.svg`}
-                alt="search icon"
-              />
+              <form className="header__menu-search-form">
+                <input
+                  className="header__menu-search-input text-body-medium-regular"
+                  type="text"
+                  placeholder="Søg blandt bibliotekets materialer"
```
spaceo, line 97: Shouldn't this be controllable with a Storybook control?

spaceo (changes requested):
Nice one :)
Have a few comments...

src/stories/autosuggest-text/autosuggest-text.scss
Adamik10: @sinejespersen perf, I adjusted it.

And yes, the same class naming will also be taken over to dpl-react, but by design we hopefully won't be creating any new css in that repo.

src/stories/autosuggest-text/AutosuggestText.stories.tsx
Adamik10: done :)

src/stories/autosuggest-text/AutosuggestText.stories.tsx
```diff
@@ -0,0 +1,54 @@
...
+  // apart from SuggestText, everything else is here just for the story context
+  // SuggestText style is directly dependent on the header and search field styling
+  <>
+    <header className="header" style={{height: "144px"}}>
```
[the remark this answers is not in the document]
Adamik10: @spaceo because it doesn't hold its height without it. It would probably need all 1 million of its children elements, which I wanted to avoid. The story context just mimics the real component, right?

src/stories/autosuggest-text/autosuggest-text.scss
```diff
@@ -0,0 +1,28 @@
...
+    display: flex;
+    align-items: center;
+    padding: 0 23px;
+
+    &:hover {
+      background-color: $c-global-secondary;
+      cursor: pointer;
+    }
+  }
+}
```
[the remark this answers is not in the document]
Adamik10: 👍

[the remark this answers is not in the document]
Adamik10: 👍

src/stories/header/Header.tsx
Adamik10: yeps 👍

PR #58, opened 2022-06-01
Autosuggestion text

Adamik10:
### **Description**

The **autosuggest text component** is a part of the autosuggest dropdown as one of three separate parts:
- autosuggest text
- autosuggest material
- autosuggest topic

This component's styling is closely bound to the search bar and header, which it depends on for styling. Thus I include a dummy search bar and header in the story. This way the _autosuggest text_ component doesn't contain anything extra/unnecessary apart from itself, but the story shows it in the context with proper styling. 

### **Screenshots**
image
image

### **Comments/Questions**


Adamik10:
all reviews done on the old PR (check comments above)

PR #60, opened 2022-06-02
Document process for authenticating when using the NPM package

kasperg:
This is somewhat duplicates what is already available in the GitHub
documentation but what scopes are actually required is not documented
and it can be handy to also have the direct link with the scopes 
handy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,28 @@ styles and icons.
 You can find the HTML output for a given story under the HTML tab inside
 storybook.
 
+### NPM package
+
+[The GitHub NPM package registry requires authentication if you are to access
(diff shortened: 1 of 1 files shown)


cableman:
Should we here mention that the local login information get mapped into docker: https://github.com/danskernesdigitalebibliotek/dpl-design-system/blob/main/docker-compose.yml#L11

kasperg:
I do not think that mapping is necessary in the current context.

Authentication is not needed when developing the design system itself. The project does not depend on any NPM packages in a GitHub registry. All integration with the GitHub NPM registry (publishing the package etc.) is handled via GitHub Actions.

For the React components it is relevant but that is another repo :smile:.

cableman (approved):

PR #65, opened 2022-06-03
Checkbox keyboard navigation

kasperg:


spaceo (approved):
Looks spitzenklasse 👍

Adamik10:
There are, however, 70 changes that still need to be approved in the UI Test GH action  👾

kasperg:
> There are, however, 70 changes that still need to be approved in the UI Test GH action

This occurs because we just just recently enabled Chromatic visual regression tests for this repo and thus there is no established baseline. Everything looks new to Chromatic.

If everybody agrees on this implementation then I think it is a good candidate for accepting the current state and then rolling forward with a proper baseline.

sinejespersen (approved):

PR #66, opened 2022-06-03
Adds search result item

kasperbirch1:
This PR adds a search result item component. The component shows a single material within a search result. 

You can see the search result item here:

- Design in Figma: https://www.figma.com/file/ETOZIfmgGS1HUfio57SOh7/S%C3%B8gning?node-id=4557%3A15899
- Component in Storybook: https://616ffdab9acbf5003ad5fd2b-tdxrynmjlj.chromatic.com/?path=/story/components-search-result-item--item

The component is made of many existing atoms. The primary task of the component is to position each atom correctly in relation to one another.

The component has props for each of the key data points which we expect to vary between items fx. title, authors and series. This allows users to see how text will wrap when text length increases.

There is also made a Favorite icon atom used by item component: https://616ffdab9acbf5003ad5fd2b-tdxrynmjlj.chromatic.com/?path=/story/atoms-icons--favourite


sinejespersen (changes requested):
The heart icon should be keyboard accessible, and also the component itself (it is clickable, right?).

kasperbirch1:
> The heart icon should be keyboard accessible, and also the component itself (it is clickable, right?).

@sinejespersen Tak, det er rettet nu.

sinejespersen (approved):

PR #68, opened 2022-06-14
add background color to .modal-btn-close

sinejespersen:
https://github.com/danskernesdigitalebibliotek/dpl-react/pull/31/files

^this pr is connected to the above. If there are two modals, both close-buttons are visible on top of each other. This covers the close button on the bottom.


src/styles/scss/shared.scss
```diff
@@ -23,7 +23,18 @@
...
 
+.modal-screen-reader-description {
+  position: absolute;
+  clip: rect(1px, 1px, 1px, 1px);
+  padding: 0;
+  border: 0;
+  height: 1px;
+  width: 1px;
+  overflow: hidden;
+}
```
kasperg, line 34: Please reconsider this.

As far as I understand this is a variation of the `.hide-visually` but [with additional considerations/improvements regarding accessibility](https://webaim.org/techniques/css/invisiblecontent/#offscreen). If that is the case then I suggest that we update the `.hide-visually` class to use this method instead of of having two classes that do the same thing.

If you think there is value in having a named BEM element (although these classes do not follow the BEM naming convention) to use here I would add a placeholder selector `%hide-visually` and `@extend` from that.

kasperg (changes requested):
I have a comment I would like you to look into.

Also please rebase your changes on `release/2` which has fixes so our GitHub Action can run properly again.

src/styles/scss/shared.scss
sinejespersen: I'll change this, you are right, I was not aware of the ```.hide-visually``` class.

spaceo (approved):

PR #72, opened 2022-06-15
Setup-node dep problem + failing favourite button a11y check

spaceo:
#### Link to issue

None

#### Description

All our Github Actions are failing. Probably because of:
https://github.com/npm/cli/issues/4998
Fixed it by using .nvmrc on setup-node tasks.

Also discovered failing a11y test on favourite button.

#### Checklist

### DIFF
--- a/.github/workflows/a11y.yml
+++ b/.github/workflows/a11y.yml
@@ -11,9 +11,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/setup-node@v2
+      - uses: actions/setup-node@v3
         with:
-          node-version: "16"
+          node-version-file: '.nvmrc'
       - run: npm ci
 
       - name: Compile styles
--- a/src/stories/button-favourite/ButtonFavourite.tsx
+++ b/src/stories/button-favourite/ButtonFavourite.tsx
@@ -6,7 +6,7 @@ interface ButtonFavouriteProps {
 
 export const ButtonFavourite = ({ fill }: ButtonFavouriteProps) => {
   return (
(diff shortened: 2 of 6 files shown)


.github/workflows/a11y.yml
```diff
@@ -11,9 +11,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - uses: actions/setup-node@v2
+      - uses: actions/setup-node@v3
         with:
-          node-version: "16"
+          node-version-file: '.nvmrc'
```
kasperg, line 16: :+1: Excellent.

If this is a problem for the NPM version that is bundled with specific node versions should we also ensure that we use same version in [`docker-compose.yml`](https://github.com/danskernesdigitalebibliotek/dpl-design-system/blob/main/docker-compose.yml#L5)? It currently specifies `node:16`.

This may also be a sign that we should at least add comments in `.nvmrc` and `docker-compose.yml` explaining that developers should keep versions in sync between the two files.

src/stories/button-favourite/ButtonFavourite.tsx
```diff
@@ -6,7 +6,7 @@ interface ButtonFavouriteProps {
 
 export const ButtonFavourite = ({ fill }: ButtonFavouriteProps) => {
   return (
-    <button className="button-favourite">
+    <button type="button" aria-label="Add to favourites" className="button-favourite">
```
kasperg, line 9: Consider if the text for `aria-label` should be in Danish or English. I am not sure if we have a established practice here.

kasperg (approved):
Great job finidng the cause of the NPM issue!

PR #81, opened 2022-07-05
Material App + Material Header

kasperbirch1:
This PR introduces a Material App "page" story, that contains Material Header.

The purpose of this is to create css classes that place all components in order and with correct spacing to each other.

**Mobil**
Skærmbillede 2022-07-06 kl  15 21 26
**Desktop**
Skærmbillede 2022-07-06 kl  15 21 35
**With period select**
Skærmbillede 2022-07-06 kl  15 21 39

### DIFF
--- a/src/stories/material-app/material-app.scss
+++ b/src/stories/material-app/material-app.scss
@@ -1,0 +1,3 @@
+.material-app {
+  // source can not be empty
+}
--- a/src/stories/material-header/material-header.tsx
+++ b/src/stories/material-header/material-header.tsx
@@ -10,12 +10,14 @@ interface MaterialHeaderProps {
   title: string;
   author: string;
   periodikum?: boolean;
+  ctaText?: string;
 }
 
 const MaterialHeader: React.FC<MaterialHeaderProps> = ({
   title,
   author,
   periodikum,
+  ctaText,
 }) => {
   return (
     <header className="material-header">
@@ -52,6 +54,7 @@ const MaterialHeader: React.FC<MaterialHeaderProps> = ({
             size="large"
           />
         </div>
+        {ctaText && <p className="mt-16 text-small-caption">{ctaText}</p>}
       </div>
     </header>
   );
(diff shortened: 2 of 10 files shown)


src/stories/material-app/material-app.scss
```diff
@@ -0,0 +1,3 @@
+.material-app {
+  // source can not be empty
```
Adamik10, line 2: lovely with the comment :)

src/stories/material-header/material-header.tsx
```diff
@@ -0,0 +1,82 @@
...
+          size="large"
+          tint="120"
+          animate={true}
+        />
+      </div>
+      <div className="material-header__content">
+        <ButtonFavourite />
+        <MaterialHeaderText title={title} author={author} />
+        <div className="material-header__availability-label-container">
+          <AvailabilityLabel
```
Adamik10, line 36: I guess we could just map through an array of four values like they do it in Checkbox.stories.tsx
Check that one out ;)

src/stories/material-header/material-periodikum-select.scss
```diff
@@ -0,0 +1,34 @@
+.material-periodikum-container {
```
Adamik10, line 1: Same thing with the -container as the other comment. Do we need the suffix?

src/stories/material-header/material-periodikum-select.tsx
```diff
@@ -0,0 +1,36 @@
+import React from "react";
+
+const MaterialPeriodikumSelect: React.FC = () => {
+  return (
+    <div className="text-small-caption material-periodikum-container ">
+      <div className="material-periodikum-select">
+        <label htmlFor="year">Årgang</label>
+        <div className="material-periodikum-select__border-container">
+          <select id="year">
+            <option value="">2022</option>
```
Adamik10, line 10: Can we instead loop over an object of values here?

Adamik10 (changes requested):
Good job :) I have a couple of comments.

Adamik10 (approved):
🥇

sinejespersen (approved):

PR #84, opened 2022-07-07
Feature/material description

kasperbirch1:


src/stories/material-app/material-app.tsx
```diff
@@ -7,13 +8,15 @@ export interface MaterialAppProps {
   author: string;
   periodikum?: boolean;
   ctaText?: string;
+  materialDescription?: string;
```
spaceo, line 11: Really a minor, but wouldn't it be sufficient to call it description?

src/stories/material-description/material-description.scss
```diff
@@ -0,0 +1,17 @@
+.material-description {
+  max-width: 750px;
+  padding: 16px;
+  padding-top: 60px;
+  padding-bottom: 44px;
```
spaceo, line 5: Just a note: As we talked about earlier I think we should have a general task looking into making sass spacing variables we can use in these cases.

src/stories/material-link/material-link.tsx
```diff
@@ -0,0 +1,35 @@
...
+      <span className="text-label-semibold">{title} </span>
+      {subTitle && <span>{subTitle}</span>}
+      {linkList.map((link, index) => (
+        <>
+          {" "}
+          <a href="/" className="link-tag" key={index}>
+            {link.text}
+          </a>
+        </>
+      ))}
```
spaceo, line 30: I think we should think semantic when designing the html layout (https://medium.com/adalab/the-importance-of-semantic-html-78e74fb75ff0)
So I think we should use an unordered list here.

cableman (approved):

src/stories/material-description/material-description.tsx
```diff
@@ -0,0 +1,76 @@
...
+      <h2 className="text-header-h4 pb-24">Beskrivelse</h2>
+      <p className="text-body-large ">{materialDescription}</p>
+      <div className="material-description__links mt-32">{MaterialLinks}</div>
+    </section>
+  );
+};
+
+export default MaterialDescription;
+
+const MaterialLinks = [
```
spaceo, line 22: Is there no linting that says that this variable should be defined before it is being used in the MaterialLink component?
I guess it works, but I would have thought that it did not work :)

src/stories/material-link/material-link.tsx
```diff
@@ -0,0 +1,35 @@
...
+  text: string;
+}
+
+export interface MaterialLinkProps {
+  title: string;
+  subTitle?: string;
+  linkList: MaterialLinkList[];
+}
+
+const MaterialLink: React.FC<MaterialLinkProps> = ({
```
spaceo, line 14: I think a more proper name for this would be something like:
`Series` or `SeriesLine`
It is an information about the series data on the shown material.

spaceo (changes requested):
Looks good! :)
Have some comments I would like you to look at.

src/stories/search-result-item/SearchResultItem.stories.tsx
```diff
@@ -30,13 +30,19 @@ export default {
...
     },
-    seriesNumber: {
-      control: { type: "text" },
-      defaultValue: "3",
-    },
-    series: {
-      control: { type: "text" },
-      defaultValue: "Små mennesker, store drømme",
+
+    materialLinkData: {
```
spaceo, line 34: Shouldn't it be seriesLineData?

PR #86, opened 2022-07-07
feature/disclosure

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-166
image

#### Description
This PR will adds the disclosure component, its styling and story. It follows the design guide that can be found at:
https://www.figma.com/file/ETOZIfmgGS1HUfio57SOh7/S%C3%B8gning?node-id=416%3A13756
https://www.figma.com/file/ETOZIfmgGS1HUfio57SOh7/S%C3%B8gning?node-id=444%3A10578

#### Screenshot of the result
image
image


src/stories/disclosure/Disclosure.stories.tsx
```diff
@@ -0,0 +1,39 @@
+import { ComponentMeta, ComponentStory } from "@storybook/react";
+import { withDesign } from "storybook-addon-designs";
+import { Disclosure, DisclosureProps } from "./Disclosure";
+
+export default {
+  title: "Atoms / Disclosure",
```
yepzdk, line 6: Please rename this to adapt to the Library structure.

base.scss
```diff
@@ -52,6 +52,7 @@
 @import "./src/stories/Library/material-description/material-description";
 @import "./src/stories/Library/material-series-line/material-series-line";
 
+@import "./src/stories/disclosure/disclosure";
```
yepzdk, line 55: Please place this in the Library folder

yepzdk (commented):
Would be nice with a more descriptive PR description and maybe a screenshot of what this results in.

src/stories/disclosure/Disclosure.stories.tsx
Adamik10: 👍 Right!

base.scss
Adamik10: Done 😉

src/stories/Library/disclosure/Disclosure.stories.tsx
```diff
@@ -0,0 +1,39 @@
...
+      url: "https://www.figma.com/file/ETOZIfmgGS1HUfio57SOh7/S%C3%B8gning?node-id=416%3A15008",
+    },
+  },
+} as ComponentMeta<typeof Disclosure>;
+
+const Template: ComponentStory<typeof Disclosure> = (args: DisclosureProps) => (
+  <Disclosure {...args} />
+);
+
+export const Defult = Template.bind({});
```
yepzdk, line 39: I guess it really don't matter, but there is a small typo with "Defult"?

src/stories/Library/disclosure/disclosure.scss
```diff
@@ -0,0 +1,32 @@
...
+  &__icon {
+    height: 40px;
+    width: 40px;
+    border-radius: 20px;
+    display: flex;
+    justify-content: center;
+
+    & > img {
+      width: 24px;
+      filter: invert(1);
```
yepzdk, line 25: If I understand correct, this is to change the color of a white icon?
So i'm not sure it is the best way, to have a consistant color across browsers. Also it is not obvious what color you end up with.
Since it is a svg that is used, we might be able to create a more flexible solution.

yepzdk (approved):
Very good description! So easy to understand the context with screenshots and links to figma. Thanks.

A couple of minor things I commented on, but nothing I think is in the way of approving.

RoWaBo (approved):

src/stories/Library/disclosure/Disclosure.stories.tsx
Adamik10: That is absolutely correct! I will fix that 😉

src/stories/Library/disclosure/disclosure.scss
Adamik10: Mmm yeah this was very much a solution I found because other things were not working. The problem is that the SVG is set as a source for an  tag. Which basically removes all the great features SVG files have. I tested if it works well in all the browsers we need, and it seems like it's behaving fine and consistent.

PR #92, opened 2022-07-12
Add Search Result Page and its sub components.

kasperbirch1:
This PR adds a collection of components used for search results page.

The components are shows as a page in a storybook component called search-result-page.

The primary task of the page component is to position each atom/component correctly in relation to one another.

You can see the search result page here:
Design in Figma: https://www.figma.com/file/ETOZIfmgGS1HUfio57SOh7/S%C3%B8gning?node-id=4557%3A15899
Component in Storybook: search-result-page

The component has props to change the title, link and total result number. This allows users to see how text will wrap when text length increases. It is also possible to toggle zero result to see the page if it does not succeed in finding results.

### DIFF
--- a/base.scss
+++ b/base.scss
@@ -51,6 +51,11 @@
 @import "./src/stories/Library/material-header/material-periodikum-select";
 @import "./src/stories/Library/material-description/material-description";
 @import "./src/stories/Library/material-series-line/material-series-line";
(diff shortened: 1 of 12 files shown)


sinejespersen (changes requested):
Unless there is a reason for the hardcoded colors, I think they are defined in a file you can use

sinejespersen (approved):

PR #94, opened 2022-07-12
Feature/5380 pause reservation component

sinejespersen:
Added a toggle button and a reservation pause component

image

image

image


base.scss
```diff
@@ -46,6 +46,8 @@
 @import "./src/stories/Library/Modals/modal-text/modal-text";
 @import "./src/stories/Library/Modals/modal-pause/modal-pause";
 @import "./src/stories/Library/material-page/material-page";
+@import "./src/stories/Library/toggle-button/toggle-button";
```
spaceo, line 49: Why is the toggle-button not placed under Library/Buttons?

src/stories/Library/pause-reservation/PauseReservation.tsx
```diff
@@ -0,0 +1,32 @@
...
+            20.09.2021 - 12.10.2021
+          </span>
+        )}
+        <div className="dpl-pause-reservation-component__flex__button">
+          <ToggleButton isChecked={isChecked} />
+        </div>
+      </div>
+    </div>
+  );
+};
```
spaceo, line 32: Shouldn't this file be under Blocks, since it is showing a composition which includes the toggle button?

src/stories/Library/pause-reservation/PauseReservation.stories.tsx
```diff
@@ -0,0 +1,24 @@
+import { ComponentMeta, ComponentStory } from "@storybook/react";
+import { withDesign } from "storybook-addon-designs";
+import { PauseReservation } from "./PauseReservation";
+
+export default {
+  title: "Library / Pause reservation",
+  component: PauseReservation,
+  decorators: [withDesign],
```
spaceo, line 8: Since you are using withDesign I think you should specify an url to the Figma design ;)

src/stories/Library/pause-reservation/PauseReservation.tsx
```diff
@@ -0,0 +1,32 @@
...
+export const PauseReservation = ({ isChecked }: PauseReservationProps) => {
+  return (
+    <div className="dpl-pause-reservation-component">
+      <div className="dpl-pause-reservation-component__pagefold" />
+      <div className="dpl-pause-reservation-component__flex">
+        <div className="dpl-pause-reservation-component__flex__reservation-icon">
+          <img src="icons/collection/Reservations.svg" alt="Reservations" />
+        </div>
+        <div className="dpl-pause-reservation-component__flex__text">
+          Sæt fysiske reserveringer på pause
```
spaceo, line 16: I think you should make this controllable from the story, so the end user can experiment with the string

src/stories/Library/pause-reservation/PauseReservation.tsx
```diff
@@ -0,0 +1,32 @@
...
+        </div>
+        {isChecked && (
+          <span
+            aria-label="Reservationer er sat på pause i følgende periode"
+            className="dpl-pause-reservation-component__flex__badge"
+          >
+            20.09.2021 - 12.10.2021
```
spaceo, line 23: I think you should make this controllable from the story, so the end user can experiment with the string

spaceo (changes requested):
Looks good! 👍 
I have some remarks I think you should consider

src/stories/Library/pause-reservation/PauseReservation.tsx
sinejespersen: I don't think so, as I dont think this is a standalone component, but rather something to nest in another component

spaceo (commented):
One last comment/question

PR #96, opened 2022-07-12
Feature/5387 empty list

sinejespersen:
image

### DIFF
--- a/src/stories/Library/Lists/list-empty/ListEmpty.stories.tsx
+++ b/src/stories/Library/Lists/list-empty/ListEmpty.stories.tsx
@@ -1,0 +1,29 @@
+import React from "react";
+import { withDesign } from "storybook-addon-designs";
+import { ComponentMeta, ComponentStory } from "@storybook/react";
+import ListEmpty from "./ListEmpty";
+
+export default {
+  title: "Library / Lists / List - empty",
+  component: ListEmpty,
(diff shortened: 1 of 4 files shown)


kasperbirch1 (approved):

src/stories/Library/Lists/list-empty/ListEmpty.stories.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { ComponentMeta, ComponentStory } from "@storybook/react";
+import ListEmpty from "./ListEmpty";
+
+export default {
+  title: "Library / Lists / List - empty",
+  component: ListEmpty,
+  decorators: [withDesign],
+  argTypes: {
+    text: {
+      defaultValue: "Du har i øjeblikket 0 Reserveringer",
```
cableman, line 12: Should default text be in english?

sinejespersen: They could I assume, I dont think its that important...? @spaceo perhaps has an opinion?

spaceo: I think the storybook default values should be in danish.
NB: Adam said that defaultValue will be deprecated in the future and we should use the object binded to the story component:
image

PR #102, opened 2022-07-13
5389: created fixed buttons in bottom of page

sinejespersen:
image

### DIFF
--- a/src/stories/Library/Modals/modal-loan/modal-loan.scss
+++ b/src/stories/Library/Modals/modal-loan/modal-loan.scss
@@ -47,4 +47,14 @@ $MODAL_LOAN_CONTAINER_WIDTH: 1000px;
   justify-content: flex-end;
   align-items: center;
   margin-top: 40px;
+
+  &--bottom {
+    position: fixed;
+    left: 0;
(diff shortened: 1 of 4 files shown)


src/stories/Library/Modals/modal-loan/modal-loan.scss
```diff
@@ -47,4 +47,14 @@ $MODAL_LOAN_CONTAINER_WIDTH: 1000px;
...
   align-items: center;
   margin-top: 40px;
+
+  &--bottom {
+    position: fixed;
+    left: 0;
+    bottom: 0;
+    right: 0;
+    background-color: $c-global-primary;
+    padding: 20px 0;
```
spaceo, line 57: This is just a note but I think the optimal solution to spacing properties would be to have some fixed variables/mixins for them

spaceo (approved):

src/stories/Library/Modals/modal-loan/modal-loan.scss
sinejespersen: Oh yeah, usually I remember, Ill fix it riddaway

PR #103, opened 2022-07-13
5389: style button as link

sinejespersen:
image

I restyled this, as it is not a link but a button.


spaceo (changes requested):
The PR is in general approved. We just need to talk it through how we are dealing with different types of links/buttons. Because we have similar scenarios like this PR.

PR #105, opened 2022-07-15
Rewrite SeriesLine to HorizontalTermLine

kasperbirch1:
This PR does two things

- Change the name of SeriesLine to HorizontalTermLine
The name has been changed because it is a reusable component and before it was prefix with: material-

- Change css width for material-description
Only material-description__content should have a max width

### DIFF
--- a/src/stories/Library/material-description/MaterialDescription.tsx
+++ b/src/stories/Library/material-description/MaterialDescription.tsx
@@ -1,6 +1,6 @@
-import SeriesLine from "../material-series-line/MaterialSeriesLine";
+import HorizontalTermLine from "../horizontal-term-line/HorizontalTermLine";
 
-const seriesLines = [
+const HorizontalTermLines = [
   {
     title: "Nr. 3",
     subTitle: "i serien",
@@ -54,7 +54,7 @@ const seriesLines = [
       },
     ],
   },
-].map((item) => <SeriesLine {...item} />);
+].map((item) => <HorizontalTermLine {...item} />);
 
 interface MaterialDescriptionProps {
   description?: string;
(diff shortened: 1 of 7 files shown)


src/stories/Library/material-description/MaterialDescription.tsx
```diff
@@ -1,6 +1,6 @@
-import SeriesLine from "../material-series-line/MaterialSeriesLine";
+import HorizontalTermLine from "../horizontal-term-line/HorizontalTermLine";
 
-const seriesLines = [
+const HorizontalTermLines = [
```
spaceo, line 3: Should be lower case since its a variable

src/stories/Library/material-description/MaterialDescription.tsx
```diff
@@ -66,8 +66,12 @@ const MaterialDescription: React.FC<MaterialDescriptionProps> = ({
   return (
     <section className="material-description">
       <h2 className="text-header-h4 pb-24">Beskrivelse</h2>
-      <p className="text-body-large ">{description}</p>
-      <div className="material-description__links mt-32">{seriesLines}</div>
+      <p className="text-body-large material-description__content">
+        {description}
+      </p>
+      <div className="material-description__links mt-32">
+        {HorizontalTermLines}
```
spaceo, line 73: It's a matter of taste but I think it would be more clear to have the map loop here.
I had to investigate how the variable was constructed before I understood what was happening.
And it is the approach that we most commonly use.
So something like:
```tsx
{horizontalTermLines.map((item) => <HorizontalTermLine {...item} />)}

spaceo (changes requested):

sinejespersen (approved):

PR #107, opened 2022-07-18
Feature/5389 css fixes

sinejespersen:
https://platform.dandigbib.org/issues/5389

This pr fixes some errors I've come upon after merging loan-list (and didnt catch because I didnt know how to link the dpl-react/dpl-design-system and test with current css at the time). 

Comments by files, I dont think images are necessary as these are small changes.

### DIFF
--- a/src/stories/Library/Forms/checkbox/checkbox.scss
+++ b/src/stories/Library/Forms/checkbox/checkbox.scss
@@ -54,7 +54,13 @@
   }
 
   &--hide-visually {
-    @extend .hide-visually !optional;
+    position: absolute;
+    clip: rect(1px, 1px, 1px, 1px);
+    padding: 0;
+    border: 0;
+    height: 1px;
+    width: 1px;
+    overflow: hidden;
   }
 }
 
--- a/src/stories/Library/Lists/list-materials/list-materials.scss
+++ b/src/stories/Library/Lists/list-materials/list-materials.scss
@@ -2,7 +2,7 @@
   display: flex;
   min-height: 72px;
   padding: 16px;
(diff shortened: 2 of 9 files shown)


src/stories/Library/Forms/checkbox/checkbox.scss
```diff
@@ -54,7 +54,13 @@
   }
 
   &--hide-visually {
```
sinejespersen, line 56: "@extend .hide-visually !optional;" doesnt seem to work, perhaps because of the order the scss are imported?

src/stories/Library/Lists/list-materials/list-materials.scss
```diff
@@ -22,8 +22,9 @@
   flex: 1;
 }
```
sinejespersen, line 24: This was a mistake from figma, where it would look like the selected should be styled like the disabled. The spec says otherwise though

src/stories/Library/Lists/list-reservation/list-reservation.scss
```diff
@@ -27,7 +27,9 @@ $list-reservation-space-desktop: 24px;
 }
 
 .list-reservation__note-mobile {
```
sinejespersen, line 29: all: unset breaks links (even though i stated it wouldnt :) )

src/stories/Library/Forms/checkbox/checkbox.scss
spaceo: We should have a mentally note on that. Could be a code smell.

spaceo (approved):
LGTM 👍 Please look at my comment

PR #108, opened 2022-07-25
Class for text truncation

yepzdk:

Adding a utitlity class for truncating text

Screenshot 2022-07-25 at 10 18 02


Adamik10 (approved):
A nice little class :) I don't have any comments.

PR #109, opened 2022-07-26
Modal profile menu

yepzdk:

Add user profile menu in modal with seperate components for avatar, modal header and the profile modal itself.

Screenshot 2022-07-26 at 07 06 00

### DIFF
--- a/src/stories/Library/Modals/modal-profile/ModalProfile.stories.tsx
+++ b/src/stories/Library/Modals/modal-profile/ModalProfile.stories.tsx
@@ -1,7 +1,7 @@
 import { withDesign } from "storybook-addon-designs";
 import { ComponentMeta, ComponentStory } from "@storybook/react";
 
-import { ModalProfile, ModalProfileProps } from "./ModalProfile";
+import { ModalProfile } from "./ModalProfile";
 
 export default {
   title: "Library / Modals / Profile",
--- a/src/stories/Library/Modals/modal-profile/ModalProfile.stories.tsx
+++ b/src/stories/Library/Modals/modal-profile/ModalProfile.stories.tsx
@@ -1,0 +1,66 @@
+import { withDesign } from "storybook-addon-designs";
+import { ComponentMeta, ComponentStory } from "@storybook/react";
+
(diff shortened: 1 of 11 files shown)


src/stories/Library/Modals/modal-profile/ModalProfile.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+    },
+  },
+} as ComponentMeta<typeof ModalProfile>;
+
+const Template: ComponentStory<typeof ModalProfile> = (args) => (
+  <ModalProfile {...args} />
+);
+
+export const Item = Template.bind({});
+Item.args = {
```
Adamik10, line 24: Ever considered using argTypes with a default value set? (line 10) Then it's also added as a storybook control.
E.g. in the SearchResultItem.stories.tsx
```code
argTypes: {
    heartFill: {
      name: "Heart Fill", // name can also be added here for Storybook visualisation
      control: "boolean",
      defaultValue: false,
    },
```
(https://storybook.js.org/docs/react/api/argtypes)

src/stories/Library/avatar/Avatar.stories.tsx
```diff
@@ -0,0 +1,18 @@
...
+const StoryBase: StoryBaseType<typeof Avatar> = {
+  title: "Library / Avatar",
+  decorators: [withDesign],
+  argTypes: {},
+  parameters: {
+    design: {
+      type: "figma",
+      url: "https://www.figma.com/file/xouARmJCONbzbZhpD8XpcM/Brugerprofil?node-id=1181%3A57001",
+    },
+    layout: "centered",
```
Adamik10, line 14: Nice! I didn't even know this existed. Good stuff 💯

Adamik10 (approved):
Very nice work :) Left a comment or two you can look at, but I'm approving 😉

src/stories/Library/Modals/modal-profile/ModalProfile.stories.tsx
yepzdk: Thanks. That is absolutely the right way to do it. I have updated the story.

PR #112, opened 2022-07-27
Fix link width

yepzdk:

Fix so links wont obtain the full width of its container, but instead stay the width of its content.

Before:
Screenshot 2022-07-27 at 13 20 04

After:
Screenshot 2022-07-27 at 13 19 28


Adamik10 (approved):
Hehe very nice :) Approved!

PR #113, opened 2022-07-27
Add rest of the content to material page component

Adamik10:
## Link to issue
https://reload.atlassian.net/browse/DDFSOEG-167
image

## Description
This PR adds the rest of the components to the material page app and fixes design issues of the subcomponents, as they didn't match the design brief. These fixes are described in the specific commit messages, and have been discovered during the development. 

## Screenshot of the result
Desktop:
image
image
image

Mobile:
image
image
image

## Comments and questions


src/stories/Blocks/material-manifestation-item/material-manifestation-item.scss
```diff
@@ -30,6 +32,7 @@
   display: flex;
   flex-wrap: wrap;
   gap: 6px;
+  height: fit-content;
```
RoWaBo, line 35: I don't this is relevant since you use the grid-template-rows to decide the height

RoWaBo (approved):

yepzdk (approved):
Looks really good to me.
Only one little concern about the use of a real bookcover(event though its renamed). Do we have permission to use it in a public git repository?

src/stories/Blocks/material-manifestation-item/material-manifestation-item.scss
Adamik10: You're right! It's not necessary :) I removed it now.

PR #115, opened 2022-07-29
5388: Add reusable box-shadows

yepzdk:
https://platform.dandigbib.org/issues/5388

Add shadows as utility classes for use with and without hover. Also available as scss variables.

Screenshot 2022-07-29 at 11 55 41


src/stories/Library/shadow/Shadow.tsx
```diff
@@ -0,0 +1,25 @@
...
+  "shadow-medium-hover",
+  "shadow-high-hover",
+];
+
+export const Shadow = () => {
+  return (
+    <div className="shadow-box-container m-24">
+      <p className="text-body-large">Shadows</p>
+      {shadowValues.map((shadowValue) => (
+        <div className={shadowValue}>
```
Adamik10, line 15: mmm, I think this will be throwing a React "each child must have a unique key prop" error in the console. Can you please check and potentially add a key={generateId(index)} to the div at line 15?

src/stories/Library/shadow/Shadow.tsx
```diff
@@ -0,0 +1,25 @@
+const shadowValues = [
+  "shadow-low",
+  "shadow-medium",
+  "shadow-high",
+  "shadow-low-hover",
+  "shadow-medium-hover",
+  "shadow-high-hover",
+];
+
+export const Shadow = () => {
```
Adamik10, line 10: Okay I know this is really nit-picky, but would it maybe make more sense to call this whole thing Shadows instead of Shadow? The Color variables file is also named in plural to reflect that there are more. 🙈 (also in the stories file Library / Shadows)

src/stories/Library/shadow/shadow.scss
```diff
@@ -0,0 +1,19 @@
+$shadows: "low"
+    "0px 0px 1px hsla(0deg, 0%, 0%, 0.03),  0 1px 4px hsla(0deg, 0%, 0%, 0.1), 1px 0px 2px hsla(0deg, 0%, 0%, 0.03)",
```
Adamik10, line 2: Those are some very advanced shadows 😸 love it!

Adamik10 (approved):
Great stuff! :) All good, but I do have a couple comments. If it's not too much to ask, please rename shadow into shadows to reflect the file content better and be consistent with other files.

src/stories/Library/shadow/Shadow.tsx
yepzdk: You are right. I have added a key to the map.

src/stories/Library/shadow/Shadow.tsx
yepzdk: No problem. I initially could not decide if it should be plural or not, and then looked at Tag and decided it should not. But i'm all for consistency, so I have renamed 👍

PR #116, opened 2022-07-29
5388: Add link function and hover

yepzdk:
https://platform.dandigbib.org/issues/5388

Make list-dashboard clickable and apply :hover styling.

Screenshot 2022-07-29 at 11 58 42


Adamik10 (approved):
Good stuff! Approved :) 
I don't have any comments.

PR #117, opened 2022-08-01
Merge feature/reviews into release/3

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-184

#### Description
This PR adds the review component, its story and styling. As a part of this PR, a couple more SVG files needed to be created in order to match the design. (https://www.figma.com/file/ETOZIfmgGS1HUfio57SOh7/S%C3%B8gning?node-id=1030%3A16233) A filled heart, and a grey heart were created. Normally this could be done using CSS, but we set these SVGs as sources for images, which removes this functionality.

#### Screenshot of the result
image

#### Checklist


src/stories/Library/review/review.scss
```diff
@@ -0,0 +1,36 @@
+.reviews {
+  margin: 0 $s-4xl 0 $s-4xl;
```
kasperg, line 2:
```suggestion
  margin: 0 $s-4xl;
```

src/stories/Library/review/review.scss
```diff
@@ -0,0 +1,36 @@
...
+
+  &__meta {
+    color: $c-text-secondary-gray;
+  }
+
+  &__heart {
+    // in Figma the heart is 12px but the img SVG in the browser has
+    // some white space in the top, so I went for making it visually
+    // the same height as the font to match the design brief
+    height: 14px;
```
kasperg, line 19:
```suggestion
    // Ensure SVG has same height as font despite top whitespace.
    height: 14px;
```

I appreciate the commentary but at the code level we may be able to short things down a bit.

src/stories/Library/review/Review.tsx
```diff
@@ -0,0 +1,69 @@
...
+              })}
+            </div>
+            <div className="review__headline mb-8">{headlineText}</div>
+            {bodyText && <p className="review__body mb-8">{bodyText}</p>}
+            {linkText && (
+              <a href={linkLink}>
+                <div className="link-tag text-small-caption mb-8">
+                  {linkText}
+                </div>
+              </a>
```
kasperg, line 60: Do we really need to nest a `<div>` inside an `<a>` tag here?

I would hope we could reduce it to something like this:

```suggestion
              <a href={linkLink} className="link-tag text-small-caption mb-8">{linkText}</a>
```

src/stories/Library/Forms/checkbox/checkbox.scss
```diff
@@ -22,7 +22,7 @@
   transform: translate3d(0, 0, 0);
 }
 
-.checkbox__label .checkbox__icon:first-child {
+.checkbox__label .checkbox__icon:first-of-type {
```
kasperg, line 25: Consider whether this change should be included.

Personally I do not think it makes much sense:

1. Googling has not lead me to a reason to why using `:first-child` is a problem for us.
2. We might remove it in `checkbox.scss` but nothing prevents us from adding it elsewhere tomorrow. If we want to avoid using `:first-child` altogether we should update our Stylelint configuration to raise errors for `:first-child`.
3. The change it not relevant to the issue at hand.

src/stories/Library/review/Review.tsx
```diff
@@ -0,0 +1,69 @@
...
+  linkLink,
+}) => {
+  const filledHeartsArray = Array.from({ length: hearts }, (_, i) => i + 1);
+  const emptyHeartsArray = Array.from({ length: 6 - hearts }, (_, i) => i + 1);
+  const reviewsArray = Array.from({ length: numberOfReviews }, (_, i) => i + 1);
+  return (
+    <ul className="reviews">
+      {reviewsArray.map((reviewNumber) => {
+        return (
+          <li className="review text-small-caption" key={reviewNumber}>
```
kasperg, line 29: :+1: for using list elements here.

src/stories/Library/review/Review.tsx
```diff
@@ -0,0 +1,69 @@
...
+  metaHeadline: string;
+  hearts: 1 | 2 | 3 | 4 | 5 | 6;
+  headlineText: string;
+  bodyText?: string;
+  linkText: string;
+  linkLink?: string;
+};
+
+export const Review: React.FC<ReviewProps> = ({
+  numberOfReviews,
```
kasperg, line 14: This can be a bit confusing. We have a component called Review (singular) which can render many reviews (plural). For the sake of simplicity and the design where it is important that we manage multiple reviews properly I get why you are doing this instead of creating two components (e.g. Review and ReviewList).

Another approach might be to consider how it is handled with checkboxes: https://github.com/danskernesdigitalebibliotek/dpl-design-system/blob/release/3/src/stories/Library/Forms/checkbox/Checkbox.stories.tsx.

src/stories/Library/review/review.scss
```diff
@@ -0,0 +1,36 @@
...
+    }
+  }
+
+  &__headline {
+    color: $c-text-secondary-gray;
+  }
+
+  &__body {
+    max-width: 545px;
+    line-height: 16px;
```
kasperg, line 33: Consider if we should add `.text-body-small-medium` or `.text-body-small-regular` to the element instead of specifying the line height for the review body text.

It seems only specific that we should add this here instead of using the general typography setup.

kasperg (commented):
:+1: Lots of good stuff here and some comments for you to consider.

src/stories/Blocks/status-userprofile/statusUserprofile.stories.tsx
```diff
@@ -37,6 +37,7 @@ export default {
           },
           title: "Afleveret for sent",
           showDot: true,
+          href: "/",
```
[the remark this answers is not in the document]
Adamik10: @kasperg it is from when I had to merge release/3 into my branch after resolving merge conflicts. Afterwards the first 2 GH actions started failing:
image
It was typescript needing the href prop for one component that wasn't getting it. Question is - how come it made it all the way to my branch? That would mean that code with this error got merged into release/3 🤔 food for thoughts

src/stories/Library/review/Review.tsx
```diff
@@ -0,0 +1,69 @@
...
+        return (
+          <li className="review text-small-caption" key={reviewNumber}>
+            <div className="review__meta mb-8">{metaHeadline}</div>
+            <div className="mb-4">
+              {filledHeartsArray.map((value) => {
+                return (
+                  <img
+                    key={value}
+                    src="icons/basic/icon-heart-filled.svg"
+                    alt="icon-heart-filled"
```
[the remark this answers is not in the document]
Adamik10: Great point! Also wonderful that you sent me that link with explanation! 💯 🥇

public/icons/basic/icon-heart-grey.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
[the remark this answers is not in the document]
Adamik10: Yes, this one has been added by me using a different stroke color, as when it's used as an img src, we cannot adjust the coloring.

src/stories/Library/review/Review.tsx
```diff
@@ -0,0 +1,69 @@
...
+}) => {
+  const filledHeartsArray = Array.from({ length: hearts }, (_, i) => i + 1);
+  const emptyHeartsArray = Array.from({ length: 6 - hearts }, (_, i) => i + 1);
+  const reviewsArray = Array.from({ length: numberOfReviews }, (_, i) => i + 1);
+  return (
+    <ul className="reviews">
+      {reviewsArray.map((reviewNumber) => {
+        return (
+          <li className="review text-small-caption" key={reviewNumber}>
+            <div className="review__meta mb-8">{metaHeadline}</div>
```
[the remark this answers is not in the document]
Adamik10: Yes yes yes :) makes sense!

src/stories/Library/review/Review.tsx
```diff
@@ -0,0 +1,69 @@
...
+                  <img
+                    key={value}
+                    src="icons/basic/icon-heart-grey.svg"
+                    alt="icon-heart-grey"
+                    className="review__heart"
+                  />
+                );
+              })}
+            </div>
+            <div className="review__headline mb-8">{headlineText}</div>
```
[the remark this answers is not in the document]
Adamik10: Mmmm I would actually disagree here - we could possibly use h6, but then we would also need to adjust the css to overrule all the predetermined styling that comes with an h6 by default, which we don't want - all the styling we need comes from the parent's class. 
I would even go as far as to say that this is not really a headline. I just called it that because it was the closest thing to a headline I can think of. It's a shorter blob of text above a longer blob of text:
image

But if you still have a different opinion, or have any other concerns in regards to this that I have overseen, I'm listening :) (like I don't know - google will like the site better if we use an h6 or something like that)

src/stories/Library/review/review.scss
Adamik10: 👍

src/stories/Library/review/review.scss
Adamik10: 👍 ❤️

src/stories/Library/review/Review.tsx
Adamik10: You are absolutely right! I was overcomplicating it for myself hehe :)

src/stories/Library/Forms/checkbox/checkbox.scss
Adamik10: I understand your thinking, but I really wanted to get rid of this nasty red error from the console:
image

src/stories/Library/review/Review.tsx
Adamik10: Thanks for inspiration :) I adjusted it similarly.

src/stories/Library/review/review.scss
Adamik10: Mmm well the general typography is not consistent with the design brief. The block of text is already inheriting  .text-small-caption as specified that it should. All fits but the line height, so I adjusted it for the specific block of text. It is defined a bit weirdly in the typography classes as 120%, which just doesn't match. And upon testing it out with and without adjustments - it is easier to read for sure when we follow the design brief.
And I didn't want to go down the road of adjusting the predefined class 😅 

But good spotted and great point!

yepzdk (approved):
Looks good to me. Great work Adam.

PR #118, opened 2022-08-03
Ensure second menu level is always full height

kasperg:
Despite the grid layout of the header which ensures that first and
second level are always full height we need to ensure that this is
the case for the second level. We mount our search app into the menu
and thus the mount point is injected between .header__menu and 
.header__menu-second.

The mount point will then get the full height but it will be missing
for .header__menu-second.

Explicitly set full height to ensure that the search bar takes up
the entire element and the search string is neatly center aligned
vertically.

Before:

Log ind | DPL CMS — (Private Browsing) 2022-08-03 16-35-24

After:

Log ind | DPL CMS — (Private Browsing) 2022-08-03 16-35-59

### DIFF
--- a/src/stories/Blocks/header/header.scss
+++ b/src/stories/Blocks/header/header.scss
@@ -130,6 +130,10 @@
   }
 }
 
+.header__menu-second {
+  height: 100%;
+}
+
 .header__menu-search {
   display: flex;
   align-items: center;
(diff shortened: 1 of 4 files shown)


Adamik10 (approved):
Perfect :) Thanks for the fixes!

PR #120, opened 2022-08-05
Rename Material to Cover

kasperbirch1:
We are changing the name because it is more descriptive.

Be aware that className is also changed so it will fail in the dpl-react repository


Adamik10 (approved):
Very nice! Finally, the component will be named the same in both repositories ;)

kasperg (approved):
:+1: LGTM

src/stories/Library/Lists/list-reservation/ListReservationsAndLoans.tsx
```diff
@@ -34,7 +34,7 @@ export const ListReservationsAndLoans = (
       >
         <div className="list-reservation__material">
           <div>
-            <Material size="small" animate url={material.image} />
+            <Cover size="small" animate url={material.image} />
```
yepzdk, line 37: Should material.image be material.cover?

src/stories/Library/material-card/MaterialCard.tsx
```diff
@@ -1,12 +1,12 @@
...
 import { ButtonUI } from "../Buttons/button-ui/ButtonUI";
-import { Material, MaterialProps } from "../material/Material";
+import { Cover, CoverProps } from "../cover/Cover";
 
 export type MaterialCardProps = {
   title: string;
   description: string;
   isLiked: boolean;
-  material: MaterialProps;
+  material: CoverProps;
```
yepzdk, line 9: Would it make sense to also rename the property material to cover?

yepzdk (approved):
Hi @kasperbirch1 
I have added a couple of questions. Minor things you can consider.
But overall good work. Approved :)

PR #121, opened 2022-08-08
Show icons on primary button in filled variant

yepzdk:
https://platform.dandigbib.org/issues/5391

This pr adresses the posibility of having the external-link button in a filled variant. 
To achieve this a new icon component for IconExternalLink is added. Also the svg path fill and stroke has been set to currentColor, to get the ability of inheriting the color property and changing it from css. This also applies for arrow small right, since it can too be a part of a filled button. 
We might consider making currentColor default for all svgs.

Final result:
Screenshot 2022-08-08 at 14 16 54

PR #124, opened 2022-08-11
Green favorite heart when filled

yepzdk:
https://platform.dandigbib.org/issues/5410

This pr changes the filled state of the favorite heart to use identity color instead of black.

Screenshot 2022-08-11 at 10 22 39
Screenshot 2022-08-11 at 10 22 25

Also this PR fixes a bug where HSL was missing in the declaration where --identity-color is defined, resulting in an empty custom property.


sinejespersen (approved):

src/stories/Library/Icons/icon-favourite/IconFavourite.tsx
```diff
@@ -10,12 +10,12 @@ export const IconFavourite = ({ fill }: IconFavouriteProps) => {
...
       className="icon-favourite"
       viewBox="0 0 24 24"
-      fill={fill ? "#000" : "none"}
+      fill={fill ? "var(--identity-color)" : "none"}
+      stroke={fill ? "var(--identity-color)" : "black"}
       xmlns="http://www.w3.org/2000/svg"
     >
       <path
         d="M11.5 20L10.2675 18.921C5.89 15.1035 3 12.5858 3 9.49591C3 6.9782 5.057 5 7.675 5C9.154 5 10.5735 5.66213 11.5 6.70845C12.4265 5.66213 13.846 5 15.325 5C17.943 5 20 6.9782 20 9.49591C20 12.5858 17.11 15.1035 12.7325 18.9292L11.5 20Z"
-        stroke="black"
```
kasperg, line 18: Based on the discussion in #121 if it would be preferable to control the color using `currentColor` and CSS.

When filled both `stroke` and `fill` would be `currentColor`. When empty the stroke would be `currentColor`. This way the `fill` prop would toggle a modifier on `.icon-favourite` e.g. `icon-favourite--filled`.

If this is not possible or you disagree then the current solution should be fine.

kasperg (approved):
:+1: I have one comment for your consideration.

src/stories/Library/Icons/icon-favourite/IconFavourite.tsx
yepzdk: @kasperg Thank you for your suggestion. You are absolutely right, that is a more clean solution. I have changed the implementation accordingly.

PR #125, opened 2022-08-18
Introduce and use a new shared class for inverting icons

Adamik10:
### Description
We have many icons used throughout the project. They are normally black,
but sometimes we need them to be white. We prefer not to set SVG's
directly into the code if we use them many places, but rather refer to
the original svg files as background sources for image tags. (If in the
future we want to change icon x to icon y we only have to exchange the
source file instead of hunting every instance in the code.) 
Thus we need this class.

### Additional comments
The chromatic test is not ✅-ed because the visual changes to the app were
not caused by this PR. ([Chromatic + cross-fork PRs i designsystemet](https://dpl.zulipchat.com/#narrow/stream/323076-CMS-NEXT/topic/Chromatic.20.2B.20cross-fork.20PRs.20i.20designsystemet))
The plan is to merge this PR in without approving these changes in Chromatic.

### DIFF
--- a/src/stories/Library/disclosure/Disclosure.tsx
+++ b/src/stories/Library/disclosure/Disclosure.tsx
@@ -22,7 +22,7 @@ export const Disclosure: React.FC<DisclosureProps> = ({
       >
         <div className="disclosure__icon bg-identity-tint-120 m-24">
           <img
-            className="disclosure__icon"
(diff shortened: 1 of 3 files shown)


kasperg (commented):
:+1: LGTM

PR #126, opened 2022-08-22
Add shared Chromatic project token to GitHub Actions

kasperg:
This way we should be able to use Chromatic across forks without
forcing contributors to add the project token as a secret. This 
should allow us to merge contributions with increased confidence.

Exposing tokens like this may not seem like a good idea but it is an
approach suggested by Chromatic:
https://www.chromatic.com/docs/custom-ci-provider#run-chromatic-on-external-forks-of-open-source-projects

Sharing project-token‘s allows contributors and others to run 
Chromatic. They’ll be able to use our snapshots. They will not be 
able to get access to our account, settings, or accept baselines. 
This is an acceptable tradeoff for us.

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -22,8 +22,4 @@ jobs:
         uses: chromaui/action@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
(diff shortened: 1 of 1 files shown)


kasperg:
I have removed the Chromatic project token from the secrets in our fork. The fact that we get changes should show that this works.

Chromatic only reports of unreviewed/new components. I have accepted them all blindly expecting that their current implementation is as expected.

cableman (approved):

PR #127, opened 2022-08-24
Autosuggest - categories

Adamik10:
### Ticket
https://reload.atlassian.net/browse/DDFSOEG-101

### Description
This PR introduces:
- boxed text
- autosuggest-categories

### Screenshots
image
image

image
image


src/stories/Library/spacing/spacing.scss
```diff
@@ -12,6 +13,7 @@ $s-215: 215;
 
 $s-xs: 4px;
 $s-sm: 8px;
+$s-smd: 10px;
```
kasperg, line 16: Consider if we should ask Dept whether this 10px size is intentional. We use a 4/8-based size system where 10 does not really fit in.

Adamik10: I think we could even go with a 12px option (8+4) - but there really is nothing between 8 and 16 pixels right now.
But I will ask in Figma :)

src/stories/Library/autosuggest-text/AutosuggestText.stories.tsx
```diff
@@ -13,6 +13,16 @@ export default {
       defaultValue: ["Item one", "Item two"],
       control: { type: "array" },
     },
+    withCategory: {
+      name: "Show category?",
+      defaultValue: false,
+      control: { type: "boolean" },
+    },
```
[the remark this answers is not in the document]
Adamik10: That's perfect! Let's get rid of unnecessary bloating. 👍 💯

yepzdk (approved):
Looks good

PR #130, opened 2022-09-02
Change name from periodikum to periodical

kasperbirch1:
Better and international naming and the same as used in dpl-react


kasperg (approved):
:+1:

cableman (approved):

PR #131, opened 2022-09-02
Add reservation component

kasperbirch1:
## Figma design
https://www.figma.com/file/ETOZIfmgGS1HUfio57SOh7/S%C3%B8gning?node-id=4560%3A32385

#### The reservation component has two sub-components. 
- **ReservationHeader**
Displays information from the selected manifest
- **ReservationForm** 
Contains a reserve button and information about the reservation.
Below is a list of selected reservations information

#### Notes
- padding will come from the modal.

#### Other changes
- reservation-header.scss has been removed and all CSS is now in reservation.scss
- added spacing of 35 px in spacing.scss

Skærmbillede 2022-09-05 kl  13 29 57


src/stories/Blocks/reservation/reservation.scss
```diff
@@ -0,0 +1,104 @@
...
+          margin-bottom: 0;
+        }
+      }
+    }
+
+    &__list {
+      display: grid;
+      gap: $s-sm;
+
+      &__item {
```
Adamik10, line 72: Consider adding a hover shadow on the card like we do on the search result page. Visually it's the same card (just with different content) so it would make sense for it to behave the same, right?

src/stories/Blocks/reservation/reservation.scss
```diff
@@ -0,0 +1,104 @@
...
+      @extend .pt-4;
+      @extend .pb-4;
+      border: 1px solid $c-global-tertiary-1;
+      display: inline-flex;
+      justify-content: center;
+      align-items: center;
+    }
+
+    &__title {
+      @extend .text-header-h2;
```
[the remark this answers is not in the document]
kasperbirch1: They are kept in css for consistency because our utils class is not used for responsive layout

src/stories/Blocks/reservation/reservation.scss
kasperbirch1: This does not appear in the design. and they are not links in the same way as search result items

sinejespersen (approved):

PR #133, opened 2022-09-05
Feature/5311 pagination

sinejespersen:
https://platform.dandigbib.org/issues/5411

- Rename search result pager to result pager
- add border:0 to list-reservation (for when the component is a button)


src/stories/Library/Lists/list-reservation/list-reservation.scss
```diff
@@ -42,6 +43,18 @@ $list-reservation-space-desktop: 24px;
...
 
+.list-reservation__hidden-explanation {
+  position: absolute;
+  clip: rect(1px, 1px, 1px, 1px);
+  padding: 0;
+  border: 0;
+  height: 1px;
+  width: 1px;
+  overflow: hidden;
+  // Should include "hide-visually", but it doesn't work, I suspect due to import order
```
spaceo, line 54: I know it is a b.... but have you looked into the issue with the loading order?
I mean we had the same problem when we started on the project, but honestly did not investigate it.
It would be nice to have solved instead of introducing hacks like this.

spaceo (commented):
Look s good. I have one question though...

src/stories/Library/Lists/list-reservation/list-reservation.scss
sinejespersen: It works if I change the order of the files. I guess it's pretty easy to fix, I am just hesitant to change the order because I suspect I might break something else? SO if it should work, the "shared.scss" should be imported before the list-reservation but after the spacing......

src/stories/Library/list-buttons/list-buttons.scss
```diff
@@ -2,6 +2,7 @@
   display: flex;
   justify-content: space-between;
   @extend %text-body-medium-medium;
+  @extend %text-body-medium-medium;
```
sinejespersen, line 5:
```suggestion

```

spaceo (approved):

PR #134, opened 2022-09-05
Fix/manifestation buttons to button elements

Adamik10:
#### Description
This PR corrects HTML semantics in regard to the availability button on edition cards on material page. It was made as a span element in the first place, but it should possess all the qualities of a button. So now it's a button.

### DIFF
--- a/src/stories/Blocks/material-manifestation-item/material-manifestation-item.scss
+++ b/src/stories/Blocks/material-manifestation-item/material-manifestation-item.scss
@@ -83,7 +83,7 @@
       align-items: center;
       justify-self: end;
     }
-    button:first-of-type {
+    .reserve-button {
       margin-top: 16px;
       max-width: 158px;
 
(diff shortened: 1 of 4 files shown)


src/stories/Blocks/material-manifestation-item/material-manifestation-item.scss
```diff
@@ -83,7 +83,7 @@
       align-items: center;
       justify-self: end;
     }
-    button {
+    button:first-of-type {
```
spaceo, line 86: Isn't it fragile to take for granted that the design is determined by the order of the buttons?
I would give it a class but if there is a strong concept/methodology behind it I am open for changing my mind :)

Adamik10: That's a fair point that I haven't thought of! :) I'll change that.

kasperg (approved):
:+1: LGTM

PR #138, opened 2022-09-12
Add ReservationSucces + ReservationError

kasperbirch1:
### Figma
https://www.figma.com/file/ETOZIfmgGS1HUfio57SOh7/S%C3%B8gning?node-id=982%3A13337

## The components are used to show whether a reservation of a material has been completed or not

Skærmbillede 2022-09-12 kl  11 30 02
Skærmbillede 2022-09-12 kl  11 30 10

### DIFF
--- a/src/stories/Blocks/reservation/reservation.scss
+++ b/src/stories/Blocks/reservation/reservation.scss
@@ -6,11 +6,11 @@
 .reservation-modal--confirm {
   display: grid;
   place-content: center;
+}
 
-  button {
-    width: 193px;
-    margin: 0 auto;
-  }
+.reservation-modal__confirm-button {
+  width: 193px;
+  margin: 0 auto;
 }
 
 .reservation-modal-header {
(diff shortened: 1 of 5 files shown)


src/stories/Blocks/reservation/reservation.scss
```diff
@@ -1,8 +1,18 @@
...
 }
 
+.reservation-modal--confirm {
+  display: grid;
+  place-content: center;
+
+  button {
+    width: 193px;
+    margin: 0 auto;
+  }
```
kasperg, line 13: Consider if it is possible to target a CSS class instead of a HTML tag.

Examples: 

- Add an `reservation-modal__button` class to the button which sets the width.
- Add a modifier to the button element e.g. `btn--wide` which sets the width.

Is `margin: 0 auto;` necessary when you use `place-content: center;`?

kasperg (changes requested):
:+1: I have a few changes for you to look into.

PR #139, opened 2022-09-14
Prevent stylelint failures

kasperg:
The task has been running for quite a while [with](https://github.com/danskernesdigitalebibliotek/dpl-design-system/actions/runs/3058616328/jobs/4935025303#step:5:41) [errors](https://github.com/danskernesdigitalebibliotek/dpl-design-system/actions/runs/3059846586/jobs/4937730327#step:5:41) when running Stylelint.

This has caused our CSS to drift away from our standards unintentionally. This issue has been increasingly problematic because developers have not run Stylelint locally as it may not be part of the local development workflow.

This PR introduces some changes to address this problem:

1. Fix running Stylelint in GitHub Actions
2. Fix our Stylelint rules to use what is intended and remove what is obsolete
3. Fix our code to comply with Stylelint and Prettier rules
4. Update our developement setup to run linting during development. This should make it easier for developers to become aware of any issues before they are pushed to GitHub

See additional commentary in the individual commit comments.

This PR may have introduced UI regressions as there are a lot of changes to existing CSS. As far as I can tell from Chromatic this is not the case.


spaceo:
Great to have the lint tightened up.
LGTM 👍

spaceo (approved):

PR #142, opened 2022-09-16
Add padding css class to modal

kasperbirch1:
#### This PR is move the reservation modal to Library / Modals 
it has been moved so that it is placed with the other modals and displays in a modal layout

- Adds ModalWrapper which is a reusable component that wraps children in a modal
-  Moves the modal CSS class from shared.scss to modal.scss
Several modals share the same padding so now there is standard padding of 80px/205px but which can easily be overwritten
Skærmbillede 2022-09-16 kl  15 42 33
Skærmbillede 2022-09-16 kl  15 42 42

##### Small others changes:
- add 215 px variable to $spacers


src/stories/Library/Modals/modal-find-on-shelf/modal-find-on-shelf.scss
```diff
@@ -4,11 +4,11 @@
...
-    padding: 0 215px $s-4xl 215px;
+    padding: $s-4xl $s-7xl;
   }
 
   &__headline {
-    margin: 72px $s-md 0 $s-md;
+    margin: 0 $s-md;
 
     @include breakpoint-m {
       margin-top: $s-4xl;
```
Adamik10, line 14: We would need to remove this margin-top here then since we introduce a padding to the modal.
I would suggest always running design system locally and comparing with Figma when doing changes like this so that we don't oversee stuff.

image

src/stories/Library/Modals/modal-find-on-shelf/modal-find-on-shelf.scss
```diff
@@ -4,11 +4,11 @@
   align-items: flex-start;
 
   @include breakpoint-m {
-    padding: 0 215px $s-4xl 215px;
+    padding: $s-4xl $s-7xl;
   }
 
   &__headline {
-    margin: 72px $s-md 0 $s-md;
```
Adamik10, line 11: Since we're moving the margins up to be padding in the modal, let's add these 72 px up in the .modal-find-on-shelf class in padding: 72px $s-md (or if there is a variable for 72, then that). Because now we are missing mobile padding/margin above the headline as seen on the picture.

image

src/stories/Library/Modals/modal-loan/ModalLoan.tsx
```diff
@@ -167,18 +168,11 @@ export const ModalLoan: React.FC<ModalLoanProps> = ({
   };
 
   return (
-    <div
-      className={`modal modal-loan modal-padding ${
-        shoulShowModal ? "modal-show" : ""
-      }`}
+    <ModalWrapper
```
Adamik10, line 171: If you go with just Modal for the component then remember to change all these <ModalWrapper > mentions ;)

src/stories/Library/spacing/spacing.scss
```diff
@@ -36,7 +36,8 @@ $spacers: (
   $s-64,
   $s-80,
   $s-96,
-  $s-112
+  $s-112,
+  $s-215
```
Adamik10, line 40: Good stuff :)

src/styles/scss/shared.scss
```diff
@@ -60,55 +60,6 @@
   }
 }
 
-.modal {
```
Adamik10, line 63: wonderful :))

Adamik10 (changes requested):
Great stuff :) I noticed some inconsistencies with the Figma design for Find on shelf modal that I commented on. And I found a component that already exists that could be merged with your ModalWrapper.

Adamik10 (changes requested):
One more thing actually - please review the first 6 UI tests stories and approve/deny them, as they were made by us :)

src/stories/Library/Modals/ModalWrapper.tsx
```diff
@@ -0,0 +1,33 @@
+import React from "react";
```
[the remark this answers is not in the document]
Adamik10: My bad, I was checking the wrong repo 🤦

Adamik10 (approved):
👍

sinejespersen (approved):

PR #149, opened 2022-10-04
Add campaign design

kasperbirch1:
### This PR Adds Campaign

The campaign consists of an image and text at a fixed height. 
it is possible to swap around the order or use ellipsis on the text with the help of modifier classes

#### be aware that there is a chromatic changes on the modal profile which is handled in PR #150

Skærmbillede 2022-10-04 kl  11 26 57
**reverse**
Skærmbillede 2022-10-04 kl  11 26 58
**ellipsis**
Skærmbillede 2022-10-07 kl  08 23 03


src/stories/Library/campaign/Campaign.tsx
```diff
@@ -0,0 +1,27 @@
...
+  reverse,
+  ellipsis,
+  title,
+}) => {
+  return (
+    <section
+      className={`campaign
+      ${reverse && "campaign--reverse"}
+      ${ellipsis && "campaign--ellipsis"}
+       `}
```
kasperg, line 19: Consider controlling class names with `clsx` like we do elsewhere in this project.

kasperg (approved):
:+1: Good stuff here. A single comment needs changing and then I have a few things for you to consider.

When the class name is changed this should be good to go from my perspective.

PR #151, opened 2022-10-04
Feature/5379 reservation materials

sinejespersen:
hide-on-mobile -> to hide a button on mobile. I would have created a modifier, but the linter complained about nesting
reservation-list-page -> wrapper for reservation list
list-reservation__material -> just small things to make it look a bit more like figma

### DIFF
--- a/src/stories/Library/list-buttons/list-buttons.scss
+++ b/src/stories/Library/list-buttons/list-buttons.scss
@@ -2,39 +2,40 @@
   display: flex;
   justify-content: space-between;
   @extend %text-body-medium-medium;
+}
 
-  &__header {
-    all: unset;
-    height: 35px;
-    position: relative;
-    padding-right: 25px;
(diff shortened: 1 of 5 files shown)


src/stories/Library/list-buttons/list-buttons.scss
```diff
@@ -29,4 +30,11 @@
       display: flex;
     }
   }
+  .hide-on-mobile {
+    display: none;
+
+    @include breakpoint-s {
+      display: flex;
+    }
+  }
```
spaceo, line 39: I am a BEM novice but if you wanted a buttons modifier for hiding on mobile wouldn't it be something like this?:

```
 &__buttons--hide-on-mobile {
    display: none;
    @include breakpoint-s {
      display: flex;
    }
  }

spaceo (approved):
Approved with one note

PR #153, opened 2022-10-17
Add FacetBrowser 

kasperbirch1:
### The following has been added: 

- FacetBrowser + FacetBrowser story  
a component that contains everything that needs to be used to display a facet browser modal

- Rewrite Tag
The tag component now follows the BEM structure and a facet prop has been added that renders the component as a button with the new --outline modifier

- Facet browser dummy data
a list of facet groups and their tags retrieved from facet-browser-dummy-data.ts

### Other changes
- Add classNames to disclosure + make icon optional
This is because we should not have an icon in the facet browser disclosure and we need to be able to add a css class that changes the padding in the disclosure header

- Add  ccs class .modal--no-padding 
Since disclosures in the facet browser must go out to the edge, we need to be able to neutralize padding on the modal

Skærmbillede 2022-10-18 kl  11 45 55
Skærmbillede 2022-10-18 kl  11 45 49


src/stories/Library/Modals/modal-facet-browser/facet-browser.scss
```diff
@@ -0,0 +1,62 @@
...
+  padding-bottom: 28px;
+  padding-right: 16px;
+  cursor: pointer;
+
+  @include breakpoint-s {
+    padding-right: 42px;
+  }
+}
+
+.facet-browser-show-results-btn {
```
Adamik10, line 60: An element inside the .facet-browser block? 
```
&__show-results-btn { ... }
```

Adamik10 (changes requested):
Good stuff :))) Got a couple of comments and questions here

src/stories/Library/disclosure/disclosure.scss
```diff
@@ -82,4 +82,11 @@
       transform: scaleY(-1);
     }
   }
+
+  &.disclosure--no-header-padding .disclosure__headline {
```
Adamik10, line 86: In this case I think it makes more sense defining a modifier on the headline element.
So something like:
```
&__headline {
     xyz
     &.disclosure__headline--no-padding {
          your code
     }
}
```

Adamik10 (approved):
Lovely! Good stuff, Kasper.
I am approving this, but I left one comment in regards to the disclosure scss. Please address it before merging 😉

sinejespersen (approved):

PR #154, opened 2022-10-18
Move logic to show modal into the Modal component

kasperbirch1:
### Move logic to show modal into the Modal component
All the logic for toggle css class .modal-show is now collected in the component. it means that changes only have to take place in one place and that the various components become more readable

### Rename and move .modal-backdrop to its own css class
Since it is not an element but a block by itself, it must have its own css class
**I have checked in [dpl-react] and it is not used there yet, so this should not cause problems**

### DIFF
--- a/src/stories/Library/Modals/Modal.tsx
+++ b/src/stories/Library/Modals/Modal.tsx
@@ -9,16 +9,17 @@ interface ModalWrapperProps {
 }
 
 const Modal = ({ children, shownModal, classNames }: ModalWrapperProps) => {
-  const [shoudShowModal, setShouldShowModal] = useState(shownModal);
+  const [shouldShowModal, setShouldShowModal] = useState(shownModal);
 
   useEffect(() => setShouldShowModal(shownModal), [shownModal]);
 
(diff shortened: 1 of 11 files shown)


Adamik10 (changes requested):
Very nice initiative! :)) 
I couldn't find anything that stood out. Great job :)

Adamik10 (approved):
✅

src/stories/Library/Modals/Modal.tsx
```diff
@@ -1,23 +1,24 @@
...
 }
 
-const Modal = ({
-  children,
-  shownModal,
-  classNames,
-  toggleModal,
-}: ModalWrapperProps) => {
+const Modal = ({ children, shownModal, classNames }: ModalWrapperProps) => {
+  const [shoudShowModal, setShouldShowModal] = useState(shownModal);
```
sinejespersen, line 12:
```suggestion
  const [shouldShowModal, setShouldShowModal] = useState(shownModal);
```

sinejespersen (approved):
there is a spelling error - if you feel like fixing it go ahead - I approved because it is just a spelling error :)

PR #158, opened 2022-10-26
Add a "campaign - without text" story, styling & a cover image for it

Adamik10:
#### Ticket
[DDFSOEG-241](https://reload.atlassian.net/browse/DDFSOEG-241)

#### Description
According to the project specs we need to implement campaigns that don't have any text, but with only an image.
The image used in the story is a free stock image, so I followed how we worked with images in the past, downloaded it and added to the project to use by local reference as opposed to an external url.

#### Screenshot of the result
image

#### Additional comments
I didn't approve 5 UI tests because they don't have anything to do with this PR (and weren't caused by it). @kasperbirch1 maybe you could look at them? I believe you know better whether they look A-OK or not 🙏


kasperbirch1 (approved):
LGTM 👍

sinejespersen (approved):

PR #160, opened 2022-10-31
Use min-width: 100% for an "image only" campaign to handle small images

Adamik10:
If the image is small, we need to use min-width together with breakpoint max-width to overwrite regular campaign's image breakpoint (where the max-width is set to 30%).


spaceo (approved):
LGTM 👍

sinejespersen (approved):

PR #161, opened 2022-11-01
Facet Line design

spaceo:
This PR contains an addition to the search result page story that shows a facet line for choosing terms and a "facet selected" line for showing the selected terms.

Issue:
https://reload.atlassian.net/browse/DDFSOEG-270

Screenshot:
image


src/stories/Library/search-result-page/search-result-page.scss
```diff
@@ -11,3 +11,41 @@
...
   }
 }
+
+.facet-line {
+  &__item {
+    display: inline-block;
+    @extend %mr-8;
+    @extend %mt-8;
+
+    .dropdown__select {
```
Adamik10, line 21: Please move these external classes into the dropdown styling as modifiers (dropdown__select and dropdown__arrows) 
That way they can be reused, live in the file corresponding to the component, and are not dependent on being inside an element with the facet-line class.

src/stories/Library/search-result-page/search-result-page.scss
```diff
@@ -11,3 +11,41 @@
...
+
+.facet-line-selected {
+  @extend %mt-16;
+
+  &__item {
+    display: inline-block;
+    @extend %mr-8;
+    @extend %mt-8;
+
+    .dropdown__select {
```
Adamik10, line 41: Same with these classes that don't match the component. We would like them to be modifiers to already existing dropdown__select and dropdown__arrows (if they differ from the regular ones).

src/stories/Library/search-result-page/SearchResultPage.tsx
```diff
@@ -88,6 +90,34 @@ const SearchResult = [
   },
 ];
 
+const facetLineItems = [
```
Adamik10, line 93: Consider moving this into its own separate file in the folder to avoid clutter in this component file.
And maybe while you're at it also the search result array.

src/stories/Library/search-result-page/FacetLineSelected.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+export interface FacetLineSelectedProps {
+  items: FacetLineItem<"term">[];
+}
+
+const FacetLineSelected: FC<FacetLineSelectedProps> = ({ items }) => {
+  return (
+    <ul className="facet-line-selected">
+      {items.map(({ title }) => (
+        <li className="facet-line-selected__item">
```
Adamik10, line 13: This class is a bit indecisive in what it would like to be? 
.facet-line is the Block
&.facet-line--selected would then need two dashes and is a modifier
&__item is correctly an element

So in this case you would need to apply these classes to the HTML list item: facet-line facet-line--selected and facet-line--selected__item

image
reference: https://csswizardry.com/2013/01/mindbemding-getting-your-head-round-bem-syntax/

I would also highly recommend going to front end masters and watching https://frontendmasters.com/courses/sass/bem-introduction/ 
It helped me a lot 😉

src/stories/Library/search-result-page/search-result-page.scss
```diff
@@ -11,3 +11,41 @@
...
+      padding: 8px 16px;
+      min-width: 200px;
+    }
+    .dropdown__arrows {
+      height: 100%;
+    }
+  }
+}
+
+.facet-line-selected {
```
Adamik10, line 33: Mmmm.. okay, so this class is out of place, because "selected" is a modifier to the facet-line class. Please refer to my other comment (from the component file) and adjust these classes.

src/stories/Library/search-result-page/FacetLineSelected.tsx
```diff
@@ -0,0 +1,23 @@
...
+          <Tag showCloseIcon hasBackground>
```
Adamik10, line 14: These "selected" tags should not have outline according to the design in Figma. Please adjust accordingly.
image

Adamik10 (changes requested):
I have a couple of smaller comments, and then a bigger one regarding the BEM structure. Hit me up if you wanna talk about it.

src/stories/Library/tag/Tag.stories.tsx
```diff
@@ -18,23 +18,25 @@ export default {
...
-
-export const Tag = Template.bind({});
+export const Tag: ComponentStory<TagProps> = (args) => (
+  <TagComp {...args}>Vi anbefaler</TagComp>
+);
 Tag.args = {
-  label: "Vi anbefaler",
-  removable: false,
+  size: "large",
+  hasBackground: true,
```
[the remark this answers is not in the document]
spaceo: 👍

src/stories/Library/search-result-page/search-result-page.scss
```diff
@@ -11,3 +11,41 @@
     gap: 16px;
   }
 }
+
+.facet-line {
```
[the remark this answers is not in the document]
spaceo: 👍

src/stories/Library/search-result-page/search-result-page.scss
spaceo: no it is its own block of  markkup and class - it is the band with facets that are selected.

src/stories/Library/search-result-page/search-result-page.scss
```diff
@@ -11,3 +11,41 @@
...
+  &__item {
+    display: inline-block;
+    @extend %mr-8;
+    @extend %mt-8;
+
+    .dropdown__select {
+      @extend %text-tags;
+      height: 100%;
+      padding: 8px 16px;
```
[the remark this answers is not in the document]
spaceo: 👍

src/stories/Library/search-result-page/FacetLineSelected.tsx
spaceo: Ok I asked Kasper about that and he told me otherwise. I'll look into that.

src/stories/Library/search-result-page/SearchResultPage.tsx
spaceo: good point 👍

src/stories/Library/search-result-page/FacetLineSelected.tsx
spaceo: I have changed it, but to be honest I would like Kasper to look at the whole Tag setup, he created those classes originally. I just cleaned up the component logic.

src/stories/Library/search-result-page/FacetLineSelected.tsx
spaceo: We went through this. Handled otherwise...

src/stories/Library/search-result-page/search-result-page.scss
spaceo: We have chosen another name

src/stories/Library/search-result-page/search-result-page.scss
spaceo: 👍

src/stories/Library/tag/Tag.tsx
```diff
@@ -2,42 +2,44 @@ import { useState } from "react";
...
-);
-
-export const Tag = ({ label, removable = false, facet }: TagProps) => {
+export const Tag = ({
+  children,
+  hasBackground = false,
+  size = "small",
+  usesCursor = true,
+  showCloseIcon = false,
+  isClickable = true,
```
kasperbirch1, line 19: Do we have to have both usesCursor + isClickable ?
or could the cursor-pointer class just come on if isClickable

Adamik10 (approved):
LGTM 👍 thanks for implementing the comments :)

src/stories/Library/tag/Tag.tsx
spaceo: By having both props we open up the possibility for having a tag that you can click on but does not change background when clicked.

sinejespersen (approved):

PR #162, opened 2022-11-10
5398: patron page and remove status.scss

sinejespersen:

- rename user-info to patron-info
- create patron-page
- refactor status container, as this has changed because the relevant data for reservations does not exist
- remove "Reserveringer pr. måned"
- remove status-"atom", because it has no value on its own
- create classes for half-inputs on desktop / full inputs on mobile
- create pincode container

image


jeppekroghitk (approved):

src/stories/Library/Modals/modal-pause/ModalPause.tsx
```diff
@@ -29,33 +29,29 @@ export const ModalPause: React.FC<ModalPauseProps> = ({
...
+            <input
+              type="date"
+              name="startDate"
+              id="startDate"
+              className="datepicker-input"
+              aria-label="startDate"
+            />
           </div>
 
           <div className="datepicker">
```
spaceo, line 44: This is is not really related to this PR. But wouldn't be an idea to make the datepicker a generic design component whit its own story? Isn't it a building block that could be used in other places besides the modal pause scope?

src/stories/Library/dropdown/dropdown.scss
```diff
@@ -8,6 +8,14 @@
   }
 }
 
+.dropdown__half-on-desktop {
```
spaceo, line 11: Is it relevant to describe what is happening when hitting the desktop breakpoint? Wouldn't: `.dropdown_desktop` do it? If, for some reason, it would be decided that it should have a third witdth in the future you would have to rename the class :)

sinejespersen: I'll rename

spaceo (approved):
LGTM 👍

PR #166, opened 2022-11-30
Feature/5389 change list reservation note

sinejespersen:
#### Link to issue

https://platform.dandigbib.org/issues/5389

#### Description

Made checkbox invisible if it is not checkable (in react, it is also disabled)
list-reservation now works without mobile/desktop added (which toggles view)

#### Screenshot of the result

image

### DIFF
--- a/src/stories/Library/Lists/list-reservation/ListReservationsAndLoans.tsx
+++ b/src/stories/Library/Lists/list-reservation/ListReservationsAndLoans.tsx
@@ -49,11 +49,11 @@ export const ListReservationsAndLoans = (
             <button
               type="button"
               aria-label="note about material"
-              className={`list-reservation__note list-reservation__note--desktop ${
-                cover.noteAsWarning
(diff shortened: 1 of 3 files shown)


src/stories/Library/Lists/list-reservation/ListReservationsAndLoans.tsx
```diff
@@ -49,7 +49,7 @@ export const ListReservationsAndLoans = (
             <button
               type="button"
               aria-label="note about material"
-              className={`list-reservation__note-desktop ${
+              className={`list-reservation__note list-reservation__note--desktop ${
```
spaceo, line 52: Please use clsx

spaceo (changes requested):
Looks good.
Meanwhile... Have some comments 🧛

spaceo (approved):

turegjorup (approved):

PR #168, opened 2022-12-19
5548: fix margin, padding and fontsizes

sinejespersen:
#### Link to issue

https://platform.dandigbib.org/issues/5548

#### Description

Fix margin between elements
Fix font size 

#### Screenshot of the result

image

### DIFF
--- a/src/stories/Library/Lists/list-reservation/list-reservation.scss
+++ b/src/stories/Library/Lists/list-reservation/list-reservation.scss
@@ -95,6 +95,9 @@ $list-reservation-space-desktop: 24px;
 
   @include breakpoint-s {
     margin-top: 0;
+    p {
+      font-size: 12px;
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


jeppekroghitk (approved):

src/stories/Library/Lists/list-reservation/list-reservation.scss
```diff
@@ -93,10 +89,15 @@ $list-reservation-space-desktop: 24px;
 
   p {
     margin-top: 2px;
+    font-size: 14px;
+    margin-bottom: 16px;
   }
```
spaceo, line 94: We have talked about using our spacing variables in general. Why not using them here?

spaceo (changes requested):
Looks good. One question though...

src/stories/Library/Lists/list-reservation/list-reservation.scss
sinejespersen: Changed

spaceo (approved):
LGTM 👍

PR #170, opened 2022-12-20
Implementing hover-based underlining for searchResultItem title

kasperbirch1:
#### This pull request adds hover-based underlining to the search result item titles.

Previously, the titles were always displayed with an underline (if there is an a tag around the title). By implementing hover-based underlining, the underlines will only be displayed when the user hovers their cursor over the title. This improves the visual appeal and user experience of the search results page.

#### Screenshot of the result
The first element is shown with a hover
Skærmbillede 2022-12-20 kl  15 59 56 (3)

#### Checklist

* The underline could not be seen in the design system, but it could be seen in dpl-react, because we have used a link tag around the title.

### DIFF
--- a/src/stories/Library/search-result-item/SearchResultItem.tsx
+++ b/src/stories/Library/search-result-item/SearchResultItem.tsx
@@ -39,7 +39,9 @@ export const SearchResultItem = ({
           )}
         </div>
 
-        <h2 className="search-result-item__title text-header-h4">{title}</h2>
+        <h2 className="search-result-item__title text-header-h4">
+          <a href="">{title}</a>
+        </h2>
         <p className="text-small-caption">{`Af ${author} (${year})`}</p>
       </div>
       <div className="search-result-item__availability">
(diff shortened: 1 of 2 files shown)


jeppekroghitk (approved):

src/stories/Library/search-result-item/SearchResultItem.tsx
```diff
@@ -39,7 +39,9 @@ export const SearchResultItem = ({
           )}
         </div>
 
-        <h2 className="search-result-item__title text-header-h4">{title}</h2>
+        <h2 className="search-result-item__title text-header-h4">
+          <a href="">{title}</a>
```
kasperg, line 43: Please consider why we need implement a separate `<a>` tag which if implemented would link to the same page as the outer link for the item.

If the only intention is to add underline to the title on hover could we implement this behavior directly on the `.search-result-item__title` class?

kasperbirch1: It was not possible for me to change the styling of the a tag within the .search-result-item__title because the user agent stylesheet for the a tag will overwrite any cascading styles from the .search-result-item__title

kasperg (approved):
:+1: LGTM

PR #172, opened 2022-12-21
5548: margin zero on deskop

sinejespersen:
#### Link to issue

https://platform.dandigbib.org/issues/5548

#### Description

Margin zero on desktop, so the author and serie title does not have space between them 

#### Screenshot of the result

image


jeppekroghitk (approved):

spaceo (approved):
LGTM 👍

PR #174, opened 2022-12-29
Unify the search result cards

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DSC-8

#### Description
The search result cards had fixed height, which didn't account for content overload. This resulted in the excess content to bleed out of the cards - in some cases even overlaying other elements on the page. 
Now we make sure that on desktop long titles get cut off after two lines, and suffixed with three dots. If there are too many availability labels present, the whole card expands in height. 
Mobile implementation stays the same - cards expand and titles are always shown in their full length. 

#### Screenshot of the result
mobile: 
image

desktop:
image

image

#### Checklist

n/a

### DIFF
--- a/src/stories/Library/search-result-item/SearchResultItem.stories.tsx
+++ b/src/stories/Library/search-result-item/SearchResultItem.stories.tsx
@@ -44,6 +44,7 @@ export default {
       },
     },
     availabilityLabels: {
+      // The control is disabled because we use stories to explore different availability label counts.
       control: { type: "null" },
       defaultValue: 3,
     },
--- a/src/stories/Library/search-result-item/SearchResultItem.stories.tsx
+++ b/src/stories/Library/search-result-item/SearchResultItem.stories.tsx
@@ -61,7 +61,7 @@ Item.args = {};
 export const ContentOverload = Template.bind({});
 ContentOverload.args = {
   title:
-    "Audrey Hepburn Audrey Hepburn Audrey Hepburn Audrey Hepburn Audrey Hepburn Audrey Hepburn Audrey Hepburn Audrey Hepburn Audrey Hepburn Audrey Hepburn Audrey Hepburn Audrey Hepburn",
+    "En roman om Jon og hans breve til sin gravide kone, da han opholdt sig i en grotte hen over vinteren og forberedte hendes ankomst og de nye tider (dansk)",
   author: "Sánchez Vegara, Amaia Arrazola, Sánchez Vegara, et al.",
   availabilityLabels: 25,
 };
--- a/src/stories/Library/search-result-item/SearchResultItem.stories.tsx
+++ b/src/stories/Library/search-result-item/SearchResultItem.stories.tsx
@@ -30,7 +30,6 @@ export default {
       control: { type: "text" },
       defaultValue: "2018",
     },
-
     horizontalTermLineData: {
       control: { type: "object" },
       defaultValue: {
@@ -44,6 +43,10 @@ export default {
         ],
       },
     },
+    availabilityLabels: {
+      control: { type: "null" },
+      defaultValue: 3,
+    },
   },
 } as ComponentMeta<typeof SearchResultItem>;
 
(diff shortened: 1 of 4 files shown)


src/stories/Library/search-result-item/SearchResultItem.stories.tsx
```diff
@@ -44,6 +43,10 @@ export default {
         ],
       },
     },
+    availabilityLabels: {
+      control: { type: "null" },
```
kasperg, line 47: I wonder if there is a reason the number of availability labels does not use a standard control 🤔 .

src/stories/Library/search-result-item/SearchResultItem.tsx
```diff
@@ -43,27 +45,17 @@ export const SearchResultItem = ({
...
-          status="available"
-        />
+        {Array(availabilityLabels)
+          .fill(0)
+          .map(() => {
+            return (
+              <AvailabilityLabel
+                manifestationType="Lydbog (cd-mp3)"
+                availability="Hjemme"
+                status="available"
```
kasperg, line 55: Consider adding variations e.g. by making every other type unavailable and using a set of predefined manifestation types.

src/stories/Library/search-result-item/search-result-item.scss
```diff
@@ -8,7 +8,7 @@
   text-decoration: none;
 
   @include breakpoint-s {
-    height: 184px;
+    min-height: 184px;
```
kasperg, line 11: Please be aware of how this works in conjunction with #176 including the vertical layout this in Chromatic.

src/stories/Library/search-result-item/search-result-item.scss
```diff
@@ -67,6 +67,12 @@
   margin-top: 10px;
 
   @include breakpoint-s {
+    display: -webkit-box;
+    -webkit-line-clamp: 2;
+    -webkit-box-orient: vertical;
     margin-top: 16px;
+    max-width: 407px;
```
kasperg, line 74: The value `407px` seems to correlate with the `grid-template-columns` on L14. I wonder if both are necessary.

Also note that Chromatic shows that the availability labels are shifted a bit to the left with this change. I wonder if that is intentional.

src/stories/Library/search-result-item/SearchResultItem.stories.tsx
```diff
@@ -53,3 +56,11 @@ const Template: ComponentStory<typeof SearchResultItem> = (args) => {
...
 export const Item = Template.bind({});
 Item.args = {};
+
+export const ContentOverload = Template.bind({});
+ContentOverload.args = {
+  title:
+    "Audrey Hepburn Audrey Hepburn Audrey Hepburn Audrey Hepburn Audrey Hepburn Audrey Hepburn Audrey Hepburn Audrey Hepburn Audrey Hepburn Audrey Hepburn Audrey Hepburn Audrey Hepburn",
+  author: "Sánchez Vegara, Amaia Arrazola, Sánchez Vegara, et al.",
+  availabilityLabels: 25,
+};
```
kasperg, line 66: :+1: for adding a Story showing a more extreme example. We have actually been provided with a [list of examples which the libraries think demonstrates extremes](https://drive.google.com/file/d/1ox1pXe6YqvNnuGlazFq7uusrfWW11I9f/view?usp=sharing). There is also [one using a very long title](https://www.aakb.dk/ting/object/870970-basis%3A52863767). Consider using that to make the example relatable.

kasperg (approved):
:+1: LGTM but a few considerations for you before this is passed along.

src/stories/Library/search-result-item/SearchResultItem.tsx
```diff
@@ -43,27 +53,19 @@ export const SearchResultItem = ({
         <p className="text-small-caption">{`Af ${author} (${year})`}</p>
       </div>
       <div className="search-result-item__availability">
```
sinejespersen, line 55: Maybe a comment explaining this?

Adamik10: Good point, thanks! :)

sinejespersen (approved):

PR #175, opened 2022-12-30
5464: add recommender to design system

sinejespersen:
#### Link to issue

https://platform.dandigbib.org/issues/5464

#### Description

Add recommender to design system
Tweak the favorites-icon-styling a bit so it works without hardcoded colors in dpl-react

#### Screenshot of the result

image

#### Checklist

### DIFF
--- a/src/stories/Library/Icons/icon-favourite/icon-favourite.scss
+++ b/src/stories/Library/Icons/icon-favourite/icon-favourite.scss
@@ -3,7 +3,7 @@
   height: 24px;
   width: 24px;
   color: $c-identity-primary;
-  stroke: $c-identity-primary;
+  stroke: $c-text-primary-black;
   fill: none;
 
   &--filled {
--- a/src/stories/Library/Icons/icon-favourite/icon-favourite.scss
+++ b/src/stories/Library/Icons/icon-favourite/icon-favourite.scss
@@ -2,6 +2,9 @@
   cursor: pointer;
   height: 24px;
   width: 24px;
+  color: $c-identity-primary;
+  stroke: $c-identity-primary;
+  fill: none;
 
   &--filled {
     color: $c-identity-primary;
--- a/src/stories/Library/Icons/icon-favourite/icon-favourite.scss
+++ b/src/stories/Library/Icons/icon-favourite/icon-favourite.scss
@@ -2,7 +2,7 @@
   cursor: pointer;
   height: 24px;
   width: 24px;
-  color: $c-identity-primary;
+  color: $c-text-primary-black;
   stroke: $c-text-primary-black;
   fill: none;
 
--- a/src/stories/Library/Icons/icon-favourite/icon-favourite.scss
+++ b/src/stories/Library/Icons/icon-favourite/icon-favourite.scss
@@ -11,3 +11,15 @@
     fill: $c-identity-primary;
   }
 }
+
+.icon-favourite--bright {
+  color: $c-text-primary-white;
+  stroke: $c-text-primary-white;
+  fill: none;
+}
+
+.icon-favourite--bright-filled {
+  stroke: $c-text-primary-white;
+  color: $c-text-primary-white;
+  fill: $c-text-primary-white;
+}
--- a/src/stories/Library/recommender/Recommender.tsx
+++ b/src/stories/Library/recommender/Recommender.tsx
@@ -1,0 +1,43 @@
+import { Cover } from "../cover/Cover";
+import { ReactComponent as SvgIcon } from "../Icons/icon-favourite/icon-favourite.svg";
+
+const Recommender: React.FC = () => {
+  return (
+    <div className="recommender">
+      <h1 className="text-header-h1">
+        Kunne du lide “Audrey Hepburn” og mangler du noget at læse?
+      </h1>
+      <div className="recommender-grid">
+        {Array.from(Array(4).keys()).map(() => (
+          <div className="recommender__grid__material">
+            <div className="button-favourite">
+              <SvgIcon
+                className={
+                  Math.random() < 0.5
+                    ? "icon-favourite icon-favourite--filled"
+                    : "icon-favourite"
+                }
+              />
+            </div>
+            <Cover
+              url="images/book_cover_3.jpg"
+              size="medium"
+              animate={false}
+              tint="120"
+            />
+            <div className="recommender__grid__material__text">
+              <div className="recommender__grid__material__text__title">
+                En bog
+              </div>
+              <div className="recommender__grid__material__text__author">
+                Af forfatter1 og forfatter2
+              </div>
+            </div>
+          </div>
+        ))}
+      </div>
+    </div>
+  );
+};
+
+export default Recommender;
--- a/src/stories/Library/recommender/Recommender.tsx
+++ b/src/stories/Library/recommender/Recommender.tsx
@@ -1,41 +1,52 @@
 import { Cover } from "../cover/Cover";
 import { ReactComponent as SvgIcon } from "../Icons/icon-favourite/icon-favourite.svg";
 
-const Recommender: React.FC = () => {
+export type RecommenderData = {
+  title: string;
+  authors: string;
+  filled: boolean;
+};
+export type RecommenderProps = {
+  recommenderData: RecommenderData[];
+  title: string;
+};
+
+const Recommender: React.FC<RecommenderProps> = ({
+  recommenderData,
+  title,
+}) => {
   return (
     <div className="recommender">
-      <h1 className="text-header-h1">
-        Kunne du lide “Audrey Hepburn” og mangler du noget at læse?
-      </h1>
-      <div className="recommender-grid">
-        {Array.from(Array(4).keys()).map(() => (
-          <div className="recommender__grid__material">
-            <div className="button-favourite">
-              <SvgIcon
-                className={
-                  Math.random() < 0.5
-                    ? "icon-favourite icon-favourite--filled"
-                    : "icon-favourite"
-                }
+      <h2 className="recommender__title text-header-h1">{title}</h2>
+      <ul className="recommender__grid">
+        {recommenderData.map(({ title: recTitle, filled, authors }) => (
+          <li className="recommender-material">
+            <div className="recommender-material__favourite">
+              <div className="button-favourite button-favourite">
+                <SvgIcon
+                  className={`icon-favourite icon-favourite--bright ${
+                    filled ? "icon-favourite--bright-filled" : ""
+                  }`}
+                />
+              </div>
+            </div>
+            <div className="recommender-material__cover-container ">
+              <Cover
+                url="images/book_cover_3.jpg"
+                size="medium"
+                animate={false}
+                tint="120"
               />
             </div>
-            <Cover
-              url="images/book_cover_3.jpg"
-              size="medium"
-              animate={false}
-              tint="120"
-            />
-            <div className="recommender__grid__material__text">
-              <div className="recommender__grid__material__text__title">
-                En bog
-              </div>
-              <div className="recommender__grid__material__text__author">
-                Af forfatter1 og forfatter2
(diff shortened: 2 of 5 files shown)


jeppekroghitk (approved):

src/stories/Library/recommender/recommender.scss
```diff
@@ -0,0 +1,83 @@
...
+.recommender__grid__material {
+  border: 1px solid $c-text-secondary-gray;
+  padding: 30px;
+  display: flex;
+  width: 100%;
+  position: relative;
+  height: 350px;
+
+  .button-favourite {
+    cursor: pointer;
```
kasperg, line 38: Please consider whether it is necessary to specify `cursor: pointer` here.

1. If we want the `.button-favourite` element to use the cursor pointer when I would expect us to want this everywhere. Not only within the Recommender. Thus this should go in [`button-favourite.scss`](https://github.com/reload/dpl-design-system/blob/main/src/stories/Library/Buttons/button-favourite/button-favourite.scss#L1-L1) instead.
2. Based on my testing in [the favourite button already has a pointer cursor](https://danskernesdigitalebibliotek.github.io/dpl-design-system/?path=/story/components-button-favourite--favourite&args=fill:false) as this is specified in [`icon-favourite`](https://github.com/reload/dpl-design-system/blob/main/src/stories/Library/Icons/icon-favourite/icon-favourite.scss#L1-L1).

src/stories/Library/recommender/recommender.scss
```diff
@@ -0,0 +1,83 @@
...
+    position: absolute;
+    top: 10px;
+    right: 10px;
```
kasperg, line 41: Please consider if this positioning can be reworked as this is something we would like to avoid with BEM.

A couple of suggestions for doing so:

1. Supporting adding additional classes to the `button-favourite` component so you can add `recommender__favourite` class with positioning.
2. Adding a wrapper `<div class="recommender__favourite">`
3. Perhaps it is even possible to do something fancy with CSS layouts.

src/stories/Library/recommender/recommender.scss
```diff
@@ -0,0 +1,83 @@
...
+      color: $c-text-primary-white;
+      stroke: $c-text-primary-white;
+      fill: none;
+
+      &--filled {
+        stroke: $c-text-primary-white;
+        color: $c-text-primary-white;
+        fill: $c-text-primary-white;
+      }
+    }
```
kasperg, line 53: Please rework this. With BEM we want to avoid this level of nesting.

I suggest implementing a new modifier for the `icon-favourite` block which uses the white color e.g. `.icon-favourite--contrast` or `.icon-favourite--bright`

src/stories/Library/recommender/recommender.scss
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  .cover-container {
+    justify-content: center;
+    width: 100%;
+  }
```
kasperg, line 59: See comment above regarding nesting of elements.

[The search result item uses a wrapper div for positioning](https://github.com/danskernesdigitalebibliotek/dpl-design-system/blob/develop/src/stories/Library/search-result-item/SearchResultItem.tsx#L26).

src/stories/Library/recommender/recommender.scss
```diff
@@ -0,0 +1,83 @@
...
+}
+
+.recommender__grid__material {
+  border: 1px solid $c-text-secondary-gray;
+  padding: 30px;
+  display: flex;
+  width: 100%;
+  position: relative;
+  height: 350px;
```
kasperg, line 35: Consider what will happen if the material has a very long title or list of authors. 

This specifies a fixed height and there does not seem to be anything which tries to tame very long titles or lists of creators.

src/stories/Library/recommender/recommender.scss
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  @include breakpoint-m {
+    &:last-of-type {
+      display: none;
+    }
+  }
+}
+
+.recommender__grid__material__text {
```
kasperg, line 68: Please rework this class name and the class names below.

It has three elements. You could break it into two blocks - something like `.recommender` and `.recommender-material`. Based on that you can create the `.recommender__meta` or `.recommender-material__meta` element. Note that if you do not break it up in two you run into problems with the material title element as the `title` is also used at the top level.

src/stories/Library/recommender/Recommender.tsx
```diff
@@ -0,0 +1,43 @@
...
+            <Cover
+              url="images/book_cover_3.jpg"
+              size="medium"
+              animate={false}
+              tint="120"
+            />
+            <div className="recommender__grid__material__text">
+              <div className="recommender__grid__material__text__title">
+                En bog
+              </div>
```
kasperg, line 31: Consider if this should use a heading element.

src/stories/Library/recommender/Recommender.tsx
```diff
@@ -0,0 +1,43 @@
...
+            </div>
+            <Cover
+              url="images/book_cover_3.jpg"
+              size="medium"
+              animate={false}
+              tint="120"
+            />
+            <div className="recommender__grid__material__text">
+              <div className="recommender__grid__material__text__title">
+                En bog
```
kasperg, line 30: Consider if the title and authors should be props which can be controlled in Storybook.

It can be nice when testing how the design behaves with different variantions of content.

src/stories/Library/Icons/icon-favourite/icon-favourite.scss
```diff
@@ -2,6 +2,9 @@
   cursor: pointer;
   height: 24px;
   width: 24px;
+  color: $c-identity-primary;
+  stroke: $c-identity-primary;
+  fill: none;
```
kasperg, line 7: Please remove this change to the base favourite icon.

As discussed in Zulip the default design if the icon should not be changed.

kasperg (changes requested):
I have quite a few comments about how this is implemented. They are primarily related to our BEM and HTML structure.

src/stories/Library/recommender/Recommender.tsx
```diff
@@ -0,0 +1,43 @@
...
+      </h1>
+      <div className="recommender-grid">
+        {Array.from(Array(4).keys()).map(() => (
+          <div className="recommender__grid__material">
+            <div className="button-favourite">
+              <SvgIcon
+                className={
+                  Math.random() < 0.5
+                    ? "icon-favourite icon-favourite--filled"
+                    : "icon-favourite"
```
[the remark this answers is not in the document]
sinejespersen: I'll fix

src/stories/Library/recommender/Recommender.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+const Recommender: React.FC = () => {
+  return (
+    <div className="recommender">
+      <h1 className="text-header-h1">
+        Kunne du lide “Audrey Hepburn” og mangler du noget at læse?
+      </h1>
+      <div className="recommender-grid">
+        {Array.from(Array(4).keys()).map(() => (
+          <div className="recommender__grid__material">
```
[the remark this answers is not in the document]
sinejespersen: I'll do so

src/stories/Library/recommender/Recommender.tsx
```diff
@@ -0,0 +1,43 @@
+import { Cover } from "../cover/Cover";
+import { ReactComponent as SvgIcon } from "../Icons/icon-favourite/icon-favourite.svg";
+
+const Recommender: React.FC = () => {
+  return (
+    <div className="recommender">
+      <h1 className="text-header-h1">
+        Kunne du lide “Audrey Hepburn” og mangler du noget at læse?
+      </h1>
```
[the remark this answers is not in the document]
sinejespersen: I'll fix

src/stories/Library/recommender/recommender.scss
```diff
@@ -0,0 +1,83 @@
+.recommender {
+  background-color: $c-global-tertiary-2;
+
+  @include breakpoint-m {
+    padding: $s-6xl;
+  }
+
+  h1 {
```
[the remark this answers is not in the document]
sinejespersen: I'll fix

src/stories/Library/recommender/recommender.scss
```diff
@@ -0,0 +1,83 @@
...
+    text-align: center;
+    padding: $s-lg;
+
+    @include breakpoint-m {
+      padding-bottom: $s-6xl;
+    }
+  }
+}
+
+.recommender-grid {
```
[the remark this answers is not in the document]
sinejespersen: I'll fix

src/stories/Library/recommender/recommender.scss
```diff
@@ -0,0 +1,83 @@
...
+  display: grid;
+  grid-template-columns: 50% 50%;
+  grid-template-rows: 50% 50%;
+
+  @include breakpoint-m {
+    grid-template-columns: 33.33% 33.33% 33.33%;
+  }
+}
+
+.recommender__grid__material {
```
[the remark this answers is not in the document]
sinejespersen: I'll fix

src/stories/Library/recommender/recommender.scss
sinejespersen: I'll fix

src/stories/Library/recommender/recommender.scss
sinejespersen: I'll fix

src/stories/Library/recommender/recommender.scss
sinejespersen: I'll fix

src/stories/Library/recommender/recommender.scss
sinejespersen: I'll fix

src/stories/Library/recommender/recommender.scss
sinejespersen: Ill fix

src/stories/Library/recommender/Recommender.tsx
sinejespersen: it is now h3

src/stories/Library/recommender/recommender.scss
```diff
@@ -0,0 +1,82 @@
...
+
+  .recommender-material__meta {
+    position: absolute;
+    left: 10px;
+    bottom: 10px;
+    width: 100%;
+    padding-right: 10px;
+  }
+
+  .recommender-material__meta__title {
```
kasperg, line 64: Please rework this class name.

It has multiple elements within a block. I suggest `.recommender-material__title`. You are allowed to nest elements within a block as long as the block constitutes a semantic element.

src/stories/Library/recommender/recommender.scss
```diff
@@ -0,0 +1,82 @@
...
+    all: unset;
```
kasperg, line 65: To me unsetting everything seemed somewhat harsh. Then I thought about it and figured it was to remove remove any default link styling. With that in mind I guess this is one way to handle this.

One additional thought was that this means that we only intend for the title to be an actual link. That creates a somewhat small hit area - especially on mobile. Then I thought the entire `.recommender-material` might be a link. However that creates problems with the favourite `<button/>` which is not allowed to be placed inside an `<a/>`. We might be able to place the `<button/>` structually outside the `<a/>` but visually on top of the element. That would be a rather large refactoring so without any specifics regarding what should be linkable and what should not so let's just go with the current implementation 😅 .

kasperg (approved):
:+1: This has turned out great! Appreciate the extra work that went into this. 

Two things I trust you to work out before this is merged:

1. There is still a class name that needs to be reworked. I
2. Chromatic has detected a change to the Recommender element where the position of the cover seems to have shifted. If this is intentional then please approve the change. Otherwise I guess theere is something to fix :smile:

PR #176, opened 2023-01-03
Set min height on search result items

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-356

#### Description

Sets min height on search result items In order to reduce cumulative layout shift.

#### Screenshot of the result

If your change affects the user interface you should include a screenshot of the result with the pull request.

#### Checklist

### DIFF
--- a/src/stories/Library/search-result-item/search-result-item.scss
+++ b/src/stories/Library/search-result-item/search-result-item.scss
@@ -9,6 +9,7 @@
   min-height: 184px;
 
   @include breakpoint-s {
+    height: 184px;
     gap: 0;
     padding: 24px;
     grid-template-columns: min-content minmax(min-content, 407px) 1fr max-content;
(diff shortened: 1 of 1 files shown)


src/stories/Library/search-result-item/search-result-item.scss
```diff
@@ -6,9 +6,9 @@
   gap: 22px 16px;
   cursor: pointer;
   text-decoration: none;
+  min-height: 184px;
 
   @include breakpoint-s {
-    height: 184px;
```
kasperg, line 11: To me it seems as if `min-height: 184px` and `height: 184px` can co-exist.  Is there any reason we do not want to force this height on specific breakpoints?

kasperg (approved):
:+1: LGTM but one thing for your consideration.

src/stories/Library/search-result-item/search-result-item.scss
kasperg: Also: As far as I can tell from Chromatic this change shifts the current layout. With `min-height: 184px` the actual height becomes 185px but before it was forced to 184px.

sinejespersen (approved):

sinejespersen (approved):

PR #179, opened 2023-01-06
Add skeleton screen css for loading state style

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFSOEG-374

#### Description

Add skeleton screen css for loading state style and implement it on a search result

#### Screenshot of the result

image

#### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,6 +82,7 @@
     "react-dom": "^17.0.2",
     "react-scripts": "4.0.3",
     "sass": "^1.53.0",
+    "skeleton-screen-css": "^1.1.0",
     "storybook-addon-designs": "^6.2.1",
     "stylelint": "^14.10.0",
     "stylelint-config-prettier": "^9.0.3",
--- a/src/stories/Library/search-result-item/SearchResultItem.stories.tsx
+++ b/src/stories/Library/search-result-item/SearchResultItem.stories.tsx
@@ -2,6 +2,7 @@ import { withDesign } from "storybook-addon-designs";
 import { ComponentMeta, ComponentStory } from "@storybook/react";
 
 import { SearchResultItem } from "./SearchResultItem";
+import { SearchResultItemSkeleton } from "./SearchResultItemSkeleton";
 
 export default {
   title: "Library / Search Result Item",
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -82,6 +82,7 @@
     "react-dom": "^17.0.2",
     "react-scripts": "4.0.3",
     "sass": "^1.53.0",
+    "skeleton-screen-css": "^1.1.0",
```
Adamik10, line 85: a very nice pure css package! Good catch 😉

src/stories/Library/search-result-item/SearchResultItem.stories.tsx
```diff
@@ -47,9 +48,12 @@ export default {
...
 } as ComponentMeta<typeof SearchResultItem>;
 
-const Template: ComponentStory<typeof SearchResultItem> = (args) => {
+export const Item: ComponentStory<typeof SearchResultItem> = (args) => {
   return <SearchResultItem {...args} />;
 };
 
-export const Item = Template.bind({});
-Item.args = {};
+export const SkeletonItem: ComponentStory<
```
Adamik10, line 55: great that you added a new story for the search result item.

Adamik10 (approved):
Very nice job Mikkel :)

src/stories/Library/search-result-item/search-result-item-skeleton.scss
```diff
@@ -0,0 +1,13 @@
+.search-result-item {
+  &.ssc {
+    grid-template-columns: min-content 1fr;
+  }
+  .ssc-square {
```
kasperg, line 5: This does not follow our BEM practices - partly intentionally because `skeleton-screen-css` does not use BEM.

This might be worth noting somewhere. It is something that would apply for all skeletons so noting it in each file does not make much sense.  One way to do so would be to explicitly disable `plugin/stylelint-bem-namics` for the file but on the other hand the current implementation does not fail liniting anyways. I wonder why that is 🤔 .

package.json
spaceo: ;)

src/stories/Library/search-result-item/SearchResultItem.stories.tsx
spaceo: ;)

sinejespersen (approved):

PR #181, opened 2023-01-09
Feature/links and boxes

sinejespersen:
#### Link to issue

https://jira.itkdev.dk/browse/CMN-448

#### Description

Create bem title for list reservation item, add arrow styling and margin to counter

#### Screenshot of the result

image

#### Checklist

### DIFF
--- a/src/stories/Library/Lists/list-reservation/list-reservation.scss
+++ b/src/stories/Library/Lists/list-reservation/list-reservation.scss
@@ -92,19 +92,6 @@ $list-reservation-space-desktop: 24px;
   padding-left: 0;
   text-align: left;
 
-  & > button {
-    all: unset;
-    @extend %text-header-h4;
-    cursor: pointer;
-
-    &:hover {
-      text-decoration: underline;
-    }
-    &:focus {
-      text-decoration: underline;
-    }
-  }
-
   p {
     margin-top: 2px;
     font-size: 14px;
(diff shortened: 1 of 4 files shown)


jeppekroghitk (approved):

src/stories/Library/Lists/list-reservation/list-reservation.scss
```diff
@@ -102,6 +107,19 @@ $list-reservation-space-desktop: 24px;
   }
 }
 
+.list-reservation__header {
+  @extend %text-header-h4;
+  padding: 0;
+  cursor: pointer;
+
+  &:hover {
```
spaceo, line 115: is it not possible to write?:
```
 &:hover,  &:focus  {

File: src/stories/Library/Lists/list-reservation/list-reservation.scss [hunk 2]
```diff
@@ -111,10 +129,28 @@ $list-reservation-space-desktop: 24px;
...
+  // Show arrow on hover
+  &:hover svg {
+    @include breakpoint-s {
+      visibility: visible;
+      opacity: 1;
+      align-self: center;
+      justify-self: end;
+    }
+  }
 }
```
spaceo, line 149: Seems like the exact same thing that is done in[ search result items](https://github.com/danskernesdigitalebibliotek/dpl-design-system/blob/f3c0e9c4b545e4064fbe7f8f19af4b69234d03df/src/stories/Library/search-result-item/search-result-item.scss#L22-L36).
How about transforming it into a mixin?

spaceo (changes requested):
Looks good. A few commentz

PR #186, opened 2023-01-20
Change cover object-fit to contain

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DSC-20

#### This pull request refactors the Cover component to handle different aspect ratios. 
The changes include:

- Changing the object-fit property from cover to contain and adapting the rest of the CSS to work with this change, including adding a $cover-width variable that's needed in some of the grid's

- Moving the type TintClassesType out of the component and declaring it at the top level of the module

- Changing the misleading name for url (cover url) in storybook

#### Checklist


src/stories/Blocks/material-banner/MaterialBanner.stories.tsx
```diff
@@ -7,7 +7,10 @@ export default {
   component: MaterialBanner,
   decorators: [withDesign],
   argTypes: {},
-  parameters: {},
+  parameters: {
+    // Notifies Chromatic to pause the animations when they finish for the specific story.
+    chromatic: { pauseAnimationAtEnd: true },
```
kasperg, line 12: This says what the configuration does but not *why* we think it is important to add/have it. The commit comment for 45c4235b7881de8c17f836b90e3caf9e81444292 mentions some of it. Perhaps that is sufficient compared to duplicating this comment across all stories. Perhaps it is something that we can define globally once?

src/stories/Library/cover/cover.scss
```diff
@@ -9,75 +9,86 @@ $MATERIAL_LARGE_DESKTOP: 216px;
...
+  margin: 0 auto;
+  object-fit: contain;
+
+  &.cover__img--animate {
+    animation: 0.3s ease-out 0s 1 CoverFadeIn;
+  }
+
+  &.cover__img--shadow {
+    box-shadow: 0 4px 40px rgb(0 0 0 / 15%);
+  }
```
kasperg, line 87: :+1: Excellent to add this add a modifier instead of something specifically for the autosuggest.

src/stories/Library/material-header/material-header.scss
```diff
@@ -21,10 +21,8 @@
       padding: 0;
     }
 
-    & .cover-container {
-      & > span.cover {
-        box-shadow: 0 4px 40px rgba(0, 0, 0, 0.2);
-      }
+    .cover__img {
+      box-shadow: 0 4px 40px rgba(0, 0, 0, 0.2);
```
kasperg, line 25: Can we remove this and add the `cover__img--shadow` instead?

There might be a slight change in the definition (0.2 vs 15%) but I wonder if difference is actually intentional.

src/stories/Library/autosuggest-material/autosuggest-material.scss
```diff
@@ -37,23 +38,14 @@
...
-    }
-
     &__content {
-      flex-flow: nowrap;
-      display: flex;
-      flex-direction: row;
+      display: grid;
+      gap: 15px;
+      grid-template-columns: 71px 1fr;
       width: 95%;
```
kasperg, line 45: Consider if this change is needed.

The original approach seems more flexible. There is no reference to specific value such as 71px.

kasperg (approved):
:+1: I have a few changes for you to consider before passing things along.

src/stories/Library/autosuggest-material/autosuggest-material.scss
kasperbirch1: This is needed because we now have different cover aspects. But I have changed the gird size for the cover to use the nearest global variable

kasperbirch1: This is needed because we now have different cover aspects. But I have changed the gird size for the cover to use the nearest global variable

jeppekroghitk (approved):

PR #191, opened 2023-01-27
Introduce local fonts in src/styles/fonts/

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-399

#### Description

To improve performance, we now define and point to local font files instead of contacting an external API. 
To follow this through, when deploying the repository we now also create a font folder that our generated css can point to.

#### Screenshot of the result
n/a

#### Checklist

n/a


spaceo (approved):
LGTM 👍

src/stories/Library/typography/typography.scss
```diff
@@ -1,4 +1,64 @@
-@import "https://fonts.googleapis.com/css2?family=Gochi+Hand&family=Lora&family=Noto+Sans+JP:wght@400;500&display=swap";
+/* noto-sans-jp-regular - latin */
```
kasperg, line 1: Consider adding a comment explaining the source of these font declarations. This will probably help in the future  if we need to add additional weights, styles or even new fonts.

The current font file names also seem infer version names (v42, v26 etc.) indicating that these are updated. Going forward we will need to update these ourselves and being aware of that tool to use here is also helpful.

jeppekroghitk (approved):

src/stories/Library/typography/typography.scss
Adamik10: Great idea!

PR #194, opened 2023-02-01
Remove default spacing in article + Add spacing in infomedia + Change padding for the search result page

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-408
https://reload.atlassian.net/browse/DDFSOEG-409 

#### This PR is remove default styling from all article elements.
it is not desired because we want to use the article for semantic reasons, not for styling. 

**The only place in the design system where this has a consequence is in Infomedia**
Therefore, I have added a new CSS class to keep the desired styling of infomedia articles

**This PR also fix padding for the search result page**

#### Checklist

This was discovered because we have used article element for semantic reasons in search-result-item in dpl-react


Adamik10 (approved):
👍

jeppekroghitk (approved):

PR #196, opened 2023-02-06
Make all full-screen modals smaller as default styling

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DSC-27

#### Description
This PR adjusts the modal styling component to by default not fill the entire screen. This change affected other stories and uses of the modal component and its content. The biggest change that I introduced is to Material Banner - because the modal is smaller, the chromatic snapshot showed it looking quite broken. I adjusted the number of items that fit in a single row for tablets/smaller screen laptops (from 4 to 3).
#### Screenshot of the result
image

#### Checklist

n/a

### DIFF
--- a/src/stories/Library/Modals/Modal.tsx
+++ b/src/stories/Library/Modals/Modal.tsx
@@ -6,9 +6,15 @@ interface ModalWrapperProps {
   shownModal: boolean;
   children?: React.ReactNode;
   classNames?: string;
+  orientation?: "right";
 }
 
-const Modal = ({ children, shownModal, classNames }: ModalWrapperProps) => {
+const Modal = ({
+  children,
+  shownModal,
+  classNames,
+  orientation,
+}: ModalWrapperProps) => {
   const [shouldShowModal, setShouldShowModal] = useState(shownModal);
 
   useEffect(() => setShouldShowModal(shownModal), [shownModal]);
@@ -29,7 +35,7 @@ const Modal = ({ children, shownModal, classNames }: ModalWrapperProps) => {
         <ModalCloseButton
           idAriaDescribedBy="describemodal"
           toggleModal={toggleModal}
-          classNames={classNames}
+          orientation={orientation}
         />
         {children}
       </div>
--- a/src/stories/Library/Modals/Modal.tsx
+++ b/src/stories/Library/Modals/Modal.tsx
@@ -19,15 +19,20 @@ const Modal = ({ children, shownModal, classNames }: ModalWrapperProps) => {
     return <ModalFallbackButton toggleModal={toggleModal} />;
 
   return (
-    <div className={clsx("modal", shouldShowModal && "modal-show", classNames)}>
-      <div className="modal__screen-reader-description" id="describemodal">
-        Denne modal dækker sidens indhold, og er en demo
+    <div className="modal-backdrop">
+      <div
+        className={clsx("modal", shouldShowModal && "modal-show", classNames)}
+      >
+        <div className="modal__screen-reader-description" id="describemodal">
+          Denne modal dækker sidens indhold, og er en demo
+        </div>
+        <ModalCloseButton
+          idAriaDescribedBy="describemodal"
+          toggleModal={toggleModal}
+          classNames={classNames}
+        />
+        {children}
       </div>
-      <ModalCloseButton
-        idAriaDescribedBy="describemodal"
-        toggleModal={toggleModal}
-      />
-      {children}
     </div>
   );
 };
--- a/src/stories/Library/Modals/ModalShared.tsx
+++ b/src/stories/Library/Modals/ModalShared.tsx
@@ -28,18 +28,19 @@ export function ModalFallbackButton({
 export function ModalCloseButton({
   toggleModal,
   idAriaDescribedBy,
-  classNames,
+  orientation,
 }: {
   toggleModal: void | (() => void);
   idAriaDescribedBy: string;
-  classNames?: string;
+  orientation?: "right";
 }) {
-  const isOffset = !(classNames && classNames.includes("modal-right"));
   return (
     <ButtonUI
       describe={idAriaDescribedBy}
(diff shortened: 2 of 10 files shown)


src/stories/Library/Modals/ModalShared.tsx
```diff
@@ -28,15 +28,18 @@ export function ModalFallbackButton({
 export function ModalCloseButton({
   toggleModal,
   idAriaDescribedBy,
+  classNames,
 }: {
   toggleModal: void | (() => void);
   idAriaDescribedBy: string;
+  classNames?: string;
 }) {
+  const isOffset = !(classNames && classNames.includes("modal-right"));
```
kasperg, line 37: Please make the orientation a prop instead of a "magic" class name which infers other changes.

src/styles/scss/shared.scss
```diff
@@ -43,6 +43,13 @@
   transition: 0.3s;
   z-index: 2;
 
+  &.modal-btn-close--offset {
+    @include breakpoint-s {
+      top: 100px;
+      right: 100px;
+    }
+  }
+
```
kasperg, line 52: Consider moving all the modal styling from `shared.scss` to `modal.scss`.

I do not see a reason why it should be part of a generic file when we have a file that deals with this area of the system.

If you do this please also include `.modal-btn-fallback` further down this file. It is outside the scope of this PR though.

src/stories/Library/Modals/modal-facet-browser/FacetBrowser.tsx
```diff
@@ -19,48 +19,46 @@ const FacetBrowser: React.FC<FacetBrowserProps> = ({
...
-      <section className="facet-browser">
-        <header className="facet-browser__header">
-          <h3 className="text-header-h3">{title}</h3>
-          {clearAll && (
-            <button className="link-tag cursor-pointer facet-browser__clear-btn">
-              {clearAll}
-            </button>
-          )}
-        </header>
+  <Modal shownModal={showModal} classNames="modal-right modal--no-padding">
```
kasperg, line 22: Consider making `modal-right` a modifier i.e. `modal--right`.

It is outside the scope of this PR though.

kasperg:
> The biggest change that I introduced is to Material Banner - because the modal is smaller, the chromatic snapshot showed it looking quite broken. I adjusted the number of items that fit in a single row for tablets/smaller screen laptops (from 4 to 3)

I am somewhat confused about whether [the Chromatic changes](https://www.chromatic.com/build?appId=616ffdab9acbf5003ad5fd2b&number=677) are in fact acceptable:

1. The material banner now has 3 rows instead of 4. This makes the story containing 8 materials look somewhat odd with rows of 3, 3 and 2 materials instead of 4 and 4. If 3 materials per row stand I think we should reduce the totalt amount of materials from 8 to 6.
2. The details item with 4 materials now has vertical scrolling and 2 rows of 3 and 1 material(s). To me it seems as if it would be possible to widen the content area within the modal to make room for 4 materials.

src/stories/Library/Modals/Modal.tsx
```diff
@@ -19,15 +19,20 @@ const Modal = ({ children, shownModal, classNames }: ModalWrapperProps) => {
     return <ModalFallbackButton toggleModal={toggleModal} />;
 
   return (
-    <div className={clsx("modal", shouldShowModal && "modal-show", classNames)}>
-      <div className="modal__screen-reader-description" id="describemodal">
-        Denne modal dækker sidens indhold, og er en demo
+    <div className="modal-backdrop">
```
kasperg, line 22: For future reference I would have preferred a different structure for the class name.

`modal-backdrop` appears as a separate block but to me it is part of the `modal` block. In that case we should have made `modal` the root element with the backdrop styling. `modal__content` would then be the element which wrapped the actual UIs.

kasperg (approved):
I found it hard to review this change.

Parts of this has to do with the BEM structure of the modal which I think can be improved but is outside of the scope of the current PR.

> The biggest change that I introduced is to Material Banner - because the modal is smaller, the chromatic snapshot showed it looking quite broken. I adjusted the number of items that fit in a single row for tablets/smaller screen laptops (from 4 to 3)

I am somewhat confused about whether [the Chromatic changes](https://www.chromatic.com/build?appId=616ffdab9acbf5003ad5fd2b&number=677) are in fact acceptable:

1. The material banner now has 3 rows instead of 4. This makes the story containing 8 materials look somewhat odd with rows of 3, 3 and 2 materials instead of 4 and 4. If 3 materials per row stand I think we should reduce the totalt amount of materials from 8 to 6.
2. The details item with 4 materials now has vertical scrolling and 2 rows of 3 and 1 material(s). To me it seems as if it would be possible to widen the content area within the modal to make room for 4 materials.

All I all I think there are some elements for you to consider and then pass this along.

src/stories/Library/Modals/modal-facet-browser/FacetBrowser.tsx
Adamik10: This would require some rewriting of the react components. I think this is a box I am not going to open.

src/stories/Library/Modals/Modal.tsx
Adamik10: I agree, but I also believe that is a bridge that was crossed a long time ago by ITK, and we approved it. On top of that rubbing me the wrong way, changing it now would mean having to change the react components too, and like the previous comment, I think that deserves its own separate task.

Adamik10:
> 

Those are all valid points Kasper, thank you. I personally think it would be a very flat cosmetic change to adjust it just for the sake of the story to make it look like they fit in one line in Chromatic. Unless I am misunderstanding the component - in reality there will always be a high chance that it will have more/less items than exactly to fill up a full row. And also, when running the app 4 items do fit into one row on desktop. It's just that Chromatic uses a viewport that's less wide to capture their screenshots.

jeppekroghitk (approved):

PR #199, opened 2023-02-09
Improved expand icon animation in 'horizontal-term-line' and 'disclosure'

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DSC-37

#### This pull request enhances the HorizontalTermLine component with the following new features:

- Limits the number of terms displayed to the value set in numberOfItemsToShow, which is set to 2.

- a toggle button has been added, allowing users to switch between displaying only the limited number of terms and all of them. The button features a toggle icon that changes according to the component's state.

**Other Changes**

- A new CSS class for the icon and a modifier to flip the arrow direction have been introduced in horizontal-term-line.

- The modifier used in DisclosureControllable has been renamed and a transition has been added.
- CSS for the default Disclosure component that uses 'details' has been added

#### Screenshot of the result
image

#### Checklist

### DIFF
--- a/src/stories/Library/disclosure/disclosure.scss
+++ b/src/stories/Library/disclosure/disclosure.scss
@@ -85,8 +85,15 @@
       margin-right: $s-xl;
     }
 
-    &-open {
+    // This is for DisclosureControllable because it's not a details element. and can't use the [open] attribute.
+    &--expanded {
+      transition: transform 0.3s ease-in-out;
       transform: scaleY(-1);
     }
   }
+}
+
+.disclosure[open] .disclosure__expand {
+  transition: transform 0.3s ease-in-out;
+  transform: scaleY(-1);
 }
--- a/src/stories/Library/horizontal-term-line/HorizontalTermLine.tsx
+++ b/src/stories/Library/horizontal-term-line/HorizontalTermLine.tsx
@@ -1,3 +1,6 @@
+import clsx from "clsx";
+import { useState } from "react";
+
 export interface HorizontalTermLineList {
   url: string;
   text: string;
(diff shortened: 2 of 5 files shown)


src/stories/Library/disclosure/disclosure.scss
```diff
@@ -85,8 +85,15 @@
       margin-right: $s-xl;
     }
 
-    &-open {
+    // This is for DisclosureControllable because it's not a details element. and can't use the [open] attribute.
+    &--expanded {
```
kasperg, line 89: :+1:

kasperg (changes requested):
Based on the Chromatic changes adding this button increases the line height of the horizontal term lines from 24 to 26px. While minor I do not think that change is intentional.

It seems as if it is added because the `button` element which has default padding in the browser.

One way to approach this would be remove this in our `reset.scss`. That might have other consequences but in general buttons should get their sizing from `button.scss`.

You could also add a new button element to the horizontal term line block.

src/stories/Library/disclosure/disclosure.scss
```diff
@@ -85,8 +85,15 @@
...
+      transition: transform 0.3s ease-in-out;
```
kasperg, line 90: :+1: Nice with a little animation

kasperg (approved):
:+1: LGTM

It is worth noting that the padding reset on button CSS results in a few layout shifts for link and favourite buttons. None of the seem to be intentional but on the other hand the current padding does not seem to be intentional either.

If we think it is important to reinstate these paddings they could be added to their respective elements e.g. `button-link` and `button-favourite`. I do not think so myself.

jeppekroghitk (approved):

PR #203, opened 2023-03-01
Fix availability label accessibility

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-430

#### Description
This PR makes sure text inside availability labels can't have a bottom margin.
We need to make sure that accessibility tools don't add a it to this text, as it would then be unreadable - hidden. We can afford to do this as there is no paragraph of text following the text inside availability labels.

#### Screenshot of the result
n/a

#### Checklist

Connected PR taking these changes to dpl-react repository: https://github.com/danskernesdigitalebibliotek/dpl-react/pull/372

### DIFF
--- a/src/stories/Library/availability-label/availability-label.scss
+++ b/src/stories/Library/availability-label/availability-label.scss
@@ -32,4 +32,11 @@
       display: inline-block;
     }
   }
+
+  & p {
+    // We need to make sure that accessibility tools don't add a bottom margin to
+    // this text, as it would then be unreadable, hidden. We can do this as there
+    // is no paragraph of text following this text inside availability labels.
+    margin-bottom: 0px !important;
+  }
 }
--- a/src/stories/Library/availability-label/availability-label.scss
+++ b/src/stories/Library/availability-label/availability-label.scss
@@ -33,7 +33,7 @@
     }
   }
 
-  & p {
+  &__text {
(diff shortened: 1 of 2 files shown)


src/stories/Library/availability-label/availability-label.scss
```diff
@@ -32,4 +32,11 @@
       display: inline-block;
     }
   }
+
+  & p {
```
kasperg, line 36: Please add a new element within the availability label e.g. `.availability-label__text` or `.availability-label__type` and `.availability-label__status` instead of targetting the generic `<p>`.

src/stories/Library/availability-label/availability-label.scss
```diff
@@ -32,4 +32,11 @@
...
+    // We need to make sure that accessibility tools don't add a bottom margin to
+    // this text, as it would then be unreadable, hidden. We can do this as there
+    // is no paragraph of text following this text inside availability labels.
+    margin-bottom: 0px !important;
```
kasperg, line 40: :+1: for commenting this.

kasperg (changes requested):
I think this change needs more work to comply with our BEM structure.

kasperg (approved):
:+1: Excellent.

jeppekroghitk (approved):

PR #204, opened 2023-03-01
Fix/autosuggest mobile wrapping

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-465

#### Description
This PR fixes autosuggest wrapping on phones + a problem where on iPhones the users would have to click the search suggestion twice in order to select it. 

#### Screenshot of the result
n/a

#### Checklist

n/a

### DIFF
--- a/src/stories/Library/autosuggest-text/autosuggest-text.scss
+++ b/src/stories/Library/autosuggest-text/autosuggest-text.scss
@@ -8,7 +8,7 @@
   z-index: 1;
 
   @include breakpoint-m {
-    padding-top: 16px;
+    padding-top: $s-md;
   }
 
   &__text {
--- a/src/stories/Library/autosuggest-text/autosuggest-text.scss
+++ b/src/stories/Library/autosuggest-text/autosuggest-text.scss
@@ -12,13 +12,16 @@
   }
 
   &__text {
-    height: 40px;
     display: flex;
     align-items: center;
 
-    &:hover {
-      background-color: $c-global-secondary;
-      cursor: pointer;
+    @include breakpoint-m {
+      height: 40px;
+
+      &:hover {
+        background-color: $c-global-secondary;
+        cursor: pointer;
+      }
     }
 
     &--highlight {


src/stories/Library/autosuggest-material/autosuggest-material.scss
```diff
@@ -30,9 +30,14 @@
...
-    &:hover {
-      background-color: $c-global-secondary;
-      cursor: pointer;
+    @include breakpoint-m {
+      // Because the main autosuggest is a prominent feature we don't want the hover
+      // effect to take effect on phones, as iPhones have a known bug that requires
+      // users to double click in order to get past hover to register a click.
+      &:hover {
+        background-color: $c-global-secondary;
+        cursor: pointer;
```
kasperg, line 39: :+1: for commenting here.

Please add a source documenting/dealing with this issue as a part of the comment. When we claim that something as major as iPhones (Mobile Safari) has a known bug then adding a source adds some credence to our claim. It may also act as a way to determine if our workaround at some point can be removed.

In general we also have other elements with both hover states and click events (e.g. buttons). I wonder if these also suffer from the same problem.

.storybook/preview.js
```diff
@@ -6,7 +6,11 @@ export const parameters = {
...
     },
   },
-  // Notifies Chromatic to pause the animations when they finish at a global level
-  // This is because Chromatic doesn't show the cover image when the animation starts with opacity 0
-  chromatic: { pauseAnimationAtEnd: true },
+  chromatic: {
+    // Notifies Chromatic to pause the animations when they finish at a global level
+    // This is because Chromatic doesn't show the cover image when the animation starts with opacity 0
+    pauseAnimationAtEnd: true,
+    // Set the viewports that Chromatic tests with globally.
```
kasperg, line 13:
```suggestion
    // Set the viewport widths that Chromatic tests with globally.
```

kasperg: Please consider if we have any reference for the values we use.

When we did not specify it we have the Chromatic default, whatever that was. Now we use 400 and 1200. Are these values based on two reference devices, do they have a relation to our breakpoints or [the devices Storybook will resize to](https://storybook.js.org/docs/react/essentials/viewport).

kasperg (approved):
This looks fine on mobile but I am unsure if this also fixes the issue on desktop. 

Desktop:
Autosuggest - Text - Multiple Line Text Suggestion ⋅ Storybook 2023-03-06 10-27-56

I have some comments for you to consider before passing this along.

Mobile:
Autosuggest - Text - Multiple Line Text Suggestion ⋅ Storybook 2023-03-06 10-31-45

PR #205, opened 2023-03-06
Add `LinkButton`

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-450

#### This PR adds a `LinkButton` component and makes some changes to improve the Button and LinkButton components in the project. 

in 'dpl-react' we needed a link style as a button therefore I have added the `LinkButton` component

I have also introduced a new ButtonIcon component to modularize the code and added a helper file to define button types, sizes, and variants. These changes make the code more readable, easier to maintain, and more scalable.

By moving the getSize() and getVariant() functions to the helper file, I have centralized the code and made it more reusable. 

#### Checklist

### DIFF
--- a/src/stories/Library/Buttons/button/LinkButton.stories.tsx
+++ b/src/stories/Library/Buttons/button/LinkButton.stories.tsx
@@ -3,7 +3,7 @@ import { withDesign } from "storybook-addon-designs";
 import { LinkButton } from "./LinkButton";
 
 export default {
-  title: "Library / Links / Linkbutton",
+  title: "Library / Buttons / Link Button",
   component: LinkButton,
   decorators: [withDesign],
   argTypes: {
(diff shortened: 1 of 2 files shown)


src/stories/Library/links/LinkButton.tsx
```diff
@@ -0,0 +1,47 @@
+import clsx from "clsx";
+import { ButtonIcon } from "../Buttons/button/ButtonIcon";
+import {
+  ButtonSize,
+  ButtonType,
+  ButtonVariant,
+  getSize,
+  getVariant,
+} from "../Buttons/button/helper";
```
kasperg, line 9: I appreciate the intent of moving these functions to helpers but now we are at a place where we import helpers across components. This is something that we normally want to avoid. 

To me this is a sign that we should reconsider whether this component should be part of the `Link` group or the `Buttons` group.

kasperg (changes requested):
:+1: but this needs a little more work.

kasperg (approved):
:+1: but remember to remove the obsolete class before merging.

jeppekroghitk (approved):

PR #206, opened 2023-03-09
Refactor headings tree structure for `Material Page`

kasperbirch1:
# DPL-REACT:  https://github.com/danskernesdigitalebibliotek/dpl-react/pull/387
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-452
https://reload.atlassian.net/browse/DDFSOEG-422
https://reload.atlassian.net/browse/DDFSOEG-423

#### This PR is refactoring the headings tree structure for the `Material Page`
To improve user navigation and content comprehension

#### it also adds a new `Heading` component to the codebase. 
The component allows for dynamic generation of the HTML tag based on the level prop provided, which can be one of four options (h2, h3, h4, or h5).

#### Screenshot of the result
Skærmbillede 2023-03-09 kl  16 02 57

image

#### Checklist

### DIFF
--- a/src/stories/Library/disclosure/Disclosure.tsx
+++ b/src/stories/Library/disclosure/Disclosure.tsx
@@ -13,7 +13,7 @@ export type DisclosureProps = {
   headingLevel: HeadingLevelType;
 };
 
-export const Disclosure: React.FC<DisclosureProps> = ({
+const Disclosure: React.FC<DisclosureProps> = ({
   headline,
   children,
   icon,
--- a/src/stories/Library/disclosure/Disclosure.tsx
+++ b/src/stories/Library/disclosure/Disclosure.tsx
@@ -1,6 +1,7 @@
 import { useState } from "react";
 import { clsx } from "clsx";
 import AvailabilityLabel from "../availability-label/AvailabilityLabel";
+import Heading, { HeadingLevelType } from "../heading/Heading";
 
 export type DisclosureProps = {
   headline: string;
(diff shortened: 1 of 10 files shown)


src/stories/Library/disclosure/Disclosure.tsx
```diff
@@ -45,13 +45,13 @@ export const Disclosure: React.FC<DisclosureProps> = ({
             />
           </div>
         )}
-        <span
-          className={`disclosure__text ${
+        <h2
```
kasperg, line 48: The disclosure story is used multiple places in the design system and it seems like a stretch to assume that h2 is the right heading level everywhere in the deisgn

I think we need some way to control this from the outside. Here is a seemingly simple way to achieve this using a prop: https://dev.to/s_aitchison/psa-stop-hard-coding-heading-levels-in-your-react-components-2ekp. Note that it probably has to be adjusted to TypeScript and our use case.

kasperg (changes requested):
Good job with the screenshot but your work on disclosures needs more thought.

jeppekroghitk (approved):

src/stories/Library/disclosure/Disclosure.tsx
```diff
@@ -9,6 +10,7 @@ export type DisclosureProps = {
   withAvailability?: boolean;
   fullWidth?: boolean;
   removeHeadlinePadding?: boolean;
+  headingLevel?: HeadingLevelType;
```
kasperg, line 13: Please make the heading level required.

[The argumentation is the same as for the React components](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/387#discussion_r1135104177).

kasperg (approved):
:+1: LGTM.

Please look into a final comment/change passing this along.

PR #209, opened 2023-03-21
Reflect changes in dpl react pr383

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-449
### 
#### This pull request is reflecting https://github.com/danskernesdigitalebibliotek/dpl-react/pull/%E2%80%A6 and contains two changes:

- The 'alt' attribute for the CheckIcon has been removed since it is not necessary for decorative images.
- The 'aria-pressed' attribute has been added to the Pagefold component to improve accessibility for users who rely on assistive technology.

#### Checklist


src/stories/Library/pagefold/Pagefold.tsx
```diff
@@ -36,10 +38,30 @@ export const Pagefold: React.FC<PagefoldProps> = (props) => {
...
+        type="button"
+        aria-pressed={status === "selected"}
+        className={classes.wrapper}
+      >
+        <div className={classes.triangle} />
+        {children}
+      </button>
+    );
+  }
+
```
spaceo, line 53: I think it is a confusing to control the type of pageFold from props especially when you also introduce the `compProps`.
In this case I would probably go the imperative way and simply create a component for each purpose. Eg.: `PageFoldButton`. And then share the logic about getting the right classes via clsx.

Btw discovered that the `baseClass`const should be used [here](https://github.com/reload/dpl-design-system/blob/e15fd7a4d829695269ff4784d1d5abee127ca9c7/src/stories/Library/pagefold/Pagefold.tsx#L36) instead of defining it again.

spaceo (changes requested):
Looks good. I have a refactoring suggestion though.

spaceo (changes requested):
I think you misunderstood my point with the button types.
The idea was to have only definition of the props like `size`, `isInheriting`and so forth.
We can also go through it together if it is not clear.

spaceo (approved):
LGTM 👍

sinejespersen (approved):

PR #210, opened 2023-03-23
Feature/Loan list additions

jeppekroghitk:
#### Link to issue

https://platform.dandigbib.org/issues/5594

#### Description

This PR adds functionality to an already used class in dpl-react, providing padding at the bottom of a modal, when fixed buttons are visible, preventing the content from being hidden behind it.
Furthermore, it removes an unnecessary margin set, as well as changes the overflow values of modals to auto instead of scroll, to prevent visible scrollbars at all times.

#### Screenshot of the result

Before:
Screenshot 2023-03-23 at 09 39 34

After:
Screenshot 2023-03-23 at 09 39 57

#### Checklist


spaceo (approved):
One comment...

sinejespersen (approved):

PR #216, opened 2023-04-25
Fix 'g' letters being cut at the bottom.

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-514

#### Description

This pull request addresses the issue where letters like 'g' are cut off at the bottom.

To fix this, I added a `min-height` property to the `.search-result-item__title` class. This prevents letters like "g" from getting cut off due to the use of `text-overflow: ellipsis;` and `overflow: hidden;`. With the inclusion of `min-height: 30px;`, the display is now correct.

#### Screenshot of the result
image
Screenshot is from dpl-react

#### Checklist

Remember to update the dpl-design-system in dpl-react

### DIFF
--- a/src/stories/Library/card-list-item/card-list-item.scss
+++ b/src/stories/Library/card-list-item/card-list-item.scss
@@ -65,5 +65,6 @@
     margin-top: 16px;
     overflow: hidden;
     text-overflow: ellipsis;
+    min-height: 30px;
   }
 }


Adamik10 (approved):
Perfecto :) You're just gonna need to fix those merge conflicts.

sinejespersen (approved):

PR #218, opened 2023-05-01
Fix/autosuggest dividers

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFSOEG-520

#### Description
This PR fixes a bug for autosuggest - material suggestions where if we got less than three items the dividers between sections would not span the full width as seen on the picture:
image

#### Screenshot of the result
Two items desktop:
image

Two items mobile:
image

One item desktop:
image

One item mobile:
image

#### Checklist

n/a

### DIFF
--- a/src/stories/Library/autosuggest-material/autosuggest-material.scss
+++ b/src/stories/Library/autosuggest-material/autosuggest-material.scss
@@ -18,6 +18,18 @@
       }
     }
 
+    &.autosuggest__material-item--two {
+      @include breakpoint-m {
+        width: calc(100% / 2);
+      }
+    }
+
+    &.autosuggest__material-item--one {
+      @include breakpoint-m {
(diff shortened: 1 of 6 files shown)


src/stories/Library/autosuggest-material/autosuggest-material.scss
```diff
@@ -18,6 +18,18 @@
...
+      @include breakpoint-m {
+        width: calc(100% / 2);
+      }
+    }
+
+    &.autosuggest__material-item--one {
+      @include breakpoint-m {
+        width: 100%;
+      }
+    }
```
spaceo, line 31: Couldn't this have been solved with grid and thereby skipping the need for more classes?

spaceo (approved):
Looks good.
I have a few questions, but nothing blocking

PR #220, opened 2023-05-03
Itkdev - 577 - Rename recommender to simple material

sinejespersen:
#### Link to issue

https://jira.itkdev.dk/browse/CMN-608

#### Description

Semantics, I think recommender-material outgrew itself from when I first named it. It is used a bunch of places, and simple material makes more sense.

#### Screenshot of the result

If your change affects the user interface you should include a screenshot of the result with the pull request.

#### Checklist


src/stories/Library/recommender/recommender.scss
yepzdk, line 1: The folder and filename should also be changed.

yepzdk (changes requested):
The stories file also needs renaming.

spaceo (approved):
LGTM 👍

PR #233, opened 2023-06-07
itkdev: menu style updates

sinejespersen:
#### Link to issue

n/a

#### Description

- Make space for notifications
- Align user icon in header
- remove row-gap

#### Screenshot of the result

n/a

#### Checklist

n/a

### DIFF
--- a/src/stories/Library/Lists/list-empty/ListEmpty.tsx
+++ b/src/stories/Library/Lists/list-empty/ListEmpty.tsx
@@ -3,11 +3,12 @@ import { Links, LinksProps } from "../../links/Links";
 interface ListEmptyProps {
   text: string;
   links?: LinksProps[];
+  className: string;
 }
 
-const ListEmpty = ({ text, links }: ListEmptyProps) => {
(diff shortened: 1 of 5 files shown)


yepzdk (approved):

src/stories/Library/Lists/list-empty/ListEmpty.tsx
```diff
@@ -3,11 +3,12 @@ import { Links, LinksProps } from "../../links/Links";
 interface ListEmptyProps {
   text: string;
   links?: LinksProps[];
+  className: string;
```
spaceo, line 6: I think it should be optional defaulting to "". In that way you are not obliged to give it a class name. That's is a quite common pattern:

```typescript
interface ListEmptyProps {
  text: string;
  links?: LinksProps[];
  className?: string;
}

const ListEmpty = ({ text, links, className="" }: ListEmptyProps) => {

spaceo (approved):
Greatness.
I have a recommendation though.

PR #243, opened 2023-08-28
Feature/advanced search

Adamik10:
#### Link to issue

n/a

#### Description

This PR introduces the advanced search stories, components, and styling.

#### Screenshot of the result

desktop:
image

mobile: 
image

#### Checklist

n/a

### DIFF
--- a/src/stories/Blocks/advanced-search/advanced-search.scss
+++ b/src/stories/Blocks/advanced-search/advanced-search.scss
@@ -22,7 +22,11 @@
     display: flex;
     flex-direction: row;
     align-items: center;
-    margin: 0 0 $s-lg $s-xl;
+    margin: 0 0 $s-md $s-xl;
+
+    @include breakpoint-m {
+      margin: 0 0 $s-lg $s-xl;
+    }
   }
 
   &__clause {
@@ -45,8 +49,12 @@
   }
 
   &__filter {
-    width: 260px;
-    margin-right: $s-lg;
+    width: 100%;
+
+    @include breakpoint-s {
+      width: 260px;
+      margin-right: $s-lg;
+    }
   }
 
   &__footer {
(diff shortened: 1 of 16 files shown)


src/stories/Blocks/advanced-search/advanced-search.scss
```diff
@@ -0,0 +1,152 @@
+.advanced-search {
+  width: 100%;
+  padding: 40px $s-md;
```
spaceo, line 3: What is you thoughts about NOT using spacing variables?

src/stories/Library/multiselect/Multiselect.stories.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { withDesign } from "storybook-addon-designs";
+import { Multiselect as MultiselectComp } from "./Multiselect";
+
+export default {
+  title: "Library / Dropdown / Multiselect",
+  component: MultiselectComp,
+  decorators: [withDesign],
+  argTypes: {
+    options: {
+      defaultValue: ["Item 1", "Item 2", "Item 3", "Item 4", "Item 5"],
```
spaceo, line 11: We should stick to one language. I think the convention is that all texts in the design system should be danish. Or am I wrong?

src/stories/Library/multiselect/Multiselect.tsx
```diff
@@ -0,0 +1,53 @@
...
+  withCaption,
+}) => {
+  return (
+    <>
+      {withCaption && (
+        <div className="multiselect__caption">Caption for multiselect</div>
+      )}
+      <div className="multiselect">
+        <div className="multiselect__selected">
+          Item 1, Item 2, Item 3, Item 4
```
spaceo, line 21: Should be danish right?

spaceo (changes requested):
I think the language texts should be aligned. Otherwise great work! 👍

src/stories/Blocks/advanced-search/advanced-search.scss
Adamik10: Well we don't have a 40px one. I think I tried to match them up if they were defined, but just opted for the values in case they weren't.

src/stories/Library/multiselect/Multiselect.stories.tsx
Adamik10: I am pretty sure it's English.
image
image

src/stories/Library/multiselect/Multiselect.tsx
Adamik10: see previous answer @spaceo

src/stories/Library/multiselect/Multiselect.stories.tsx
spaceo: Ok let's stick to english if that is what is widely used! 👍

src/stories/Blocks/advanced-search/advanced-search.scss
spaceo: I have no strong opinion about this, other than the spacing system has failed 😃 
We can talk about if it should have an overhaul.
One idea could be to consider the existing system to be deprecated and keep it, and use something new onwards.

spaceo (approved):
👍 LGTM

PR #283, opened 2023-10-06
Improved space for status label in list dashboard

JacobArrow:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-207

#### Description

The current implementation does not hide the dashboard arrow on mobile devices which gives limited space for the status label.
I allowed more space by hiding the arrow on mobile and added less line hight for the status label.

#### Screenshot of the result
Normal size mobile
image
Small size mobile
image

#### Checklist


Adamik10 (approved):
Guddie

PR #284, opened 2023-10-08
Search header dropdown

Adamik10:
This PR adjusts the z-index of the search header dropdown, so that the menu opens on top of autosuggest, if autosuggest is open too.


spaceo (approved):
Splendid 👍

PR #294, opened 2023-10-10
Group Dependabot updates for Storybook

kasperg:
#### Description

We use many packages from Storybook. These are often updated in parallel  and may have dependencies between them. Thus is makes sense to group  updates of these packages in Dependabot to ensure that we get an update  to e.g. Storybook 7 in the same pull request.


spaceo (approved):
LGTM 👍

PR #332, opened 2023-11-04
Introducing button box

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-57

#### Description

This PR  introduces the button box and sticky button box concept.
The footer box classes were created in order to accomodate the sticky renewal button on the loan page when watching it on a mobile device.
The button box classes replaces the modal-loan__buttons classes, since button box is more generic and used not only in modals.

#### Screenshot of the result

https://github.com/danskernesdigitalebibliotek/dpl-design-system/assets/998889/6f1d37a3-3635-4929-9f3c-6629d8cf641f


Adamik10 (approved):
Very nice :)) 
Although I still don't think I get what the "button-box" is from just the naming. 🙈

PR #339, opened 2023-11-16
Added hover state for buttons

JacobArrow:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-77

#### Description
Buttons have a very "dead" interaction, and we need to show the user when they hover it. This adds some opacity to filled hovered buttons, and fill to outlined hovered buttons. It also aligns the filled disabled state with the figma design

#### Screenshot of the result
FilledStateHoverArrow
FilledStateHover
DisabledStateHover
ArrowStateHover
DefaultStateHover


Dresse (approved):
LGTM 👍

PR #341, opened 2023-11-17
Improve rendering of reservation labels

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-278
https://reload.atlassian.net/browse/DDFLSBP-279

#### Description

They were just put in roughly everywhere needed in dpl-react so the different views in group-modal and reservation-list had to be adjusted.

#### Screenshot of the result

image

### DIFF
--- a/src/stories/Library/Lists/list-materials/list-materials.scss
+++ b/src/stories/Library/Lists/list-materials/list-materials.scss
@@ -1,12 +1,11 @@
 .list-materials {
   display: flex;
-  min-height: 72px;
+  min-height: 96px;
   padding: 16px;
   background-color: $c-text-primary-white;
   justify-content: space-between;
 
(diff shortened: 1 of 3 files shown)


src/stories/Library/Lists/list-materials/list-materials.scss
```diff
@@ -80,3 +79,14 @@
     display: none;
   }
 }
+
+.list-materials__arrow {
+  margin-left: $s-xl;
```
Adamik10, line 84: Love it that you're using the variables we have :)))

Adamik10 (changes requested):
Great stuff. Please have a look at all the UI tests and consider whether these changes had any unpredictable effects on other views - I believe there are over 30 that need to be reviewed.

PR #347, opened 2023-11-27
DDFFORM 6 linkskomponent liste

LasseStaus:
- Added internalLink & Searchbooks SVG
- Fixed typo in URL for figma link to IconCollection
- Added LinkWithIcon with stories & types.
- Added CSS for linkWithIcon & linkIcon
- Add temporary container(size) mixin
- Added dist.zip to gitignore

#### Link to issue
[Jira ticket](https://reload.atlassian.net/jira/software/c/projects/DDFFORM/boards/485?selectedIssue=DDFFORM-6)

[CMS PR](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/516)

#### Description

This PR adds a Link component, which can be used for display either an internal link, external link, File download or link to a search result. 

- Added new SVGS
- Add the link component
- Fixed typo in figma link for iconCollection. 
- Added temporary container mixin. Will be changed later when we agree on a general structure
- Added dist.zip to .gitignore (See questions) 

#### Screenshot of the result

image

Consider wether dist.zip in .gitignore has any other unintentional implications. 
I just wanted to not unnecesarily see this in git when using the new dapple cms+react setup. 

I don't have specifics, but since this is my first design system component, feel free to just comment on anything. 

This could be
- General JSX structure.
- Specifics to storybook.
- Specifics to the CSS.
- Specifics to icon handling
- If you in general observe something / know something that would be beneficial for me to know going forward developing on this project, then please comment with that.


src/stories/Library/link-with-icon/LinkWithIcon.tsx
```diff
@@ -0,0 +1,51 @@
...
+    "rotate-180": linkType === "internal",
+  });
+
+  return (
+    <a
+      href={href}
+      target={target}
+      rel={target === "_blank" ? "noopener noreferrer" : undefined}
+      className={clsx(
+        "link-with-icon text-body-medium-regular arrow__hover--right-small"
```
kasperg, line 41: Please reconsider the class structure here.

If we are moving away from utility classes then I would as a minimum expect `text-body-medium-regular` to be included in `link-with-icon` instead of added as a separate class.

I do not have a strong opinion on whether it is OK to add `arrow__hover--right-small` separately or whether that should also be included in `link-with-icon`.

src/stories/Library/link-with-icon/link-with-icon.scss
```diff
@@ -0,0 +1,60 @@
...
+    &:hover > svg {
+      visibility: visible;
+    }
+  }
+
+  &.arrow__hover--right-small {
+    .arrow__body,
+    .arrow__head {
+      transition: none;
+    }
```
kasperg, line 42: Please reconsider this CSS.

When using BEM we do not want our blocks to interfere with eachother. *If* we want to disable animation for an arrow we should instead create a new modifier for `arrow` (e.g. `arrow--no-animation`) , disable transitions for that. Then we can use this modifier in our template for links with icons.

Also: Is ther a reason why we want to disable the default arrow animation here? I cannot tell from code or commit comments.

src/stories/Library/link-with-icon/link-with-icon.scss
```diff
@@ -0,0 +1,60 @@
...
+  }
+}
+
+.link-icon {
```
kasperg, line 46: Please reconsider this CSS.

When working with a BEM structure we want blocks or elements. If this is an element within in the `link-with-icon` block then it should be named as such e.g. `link-with-icon__icon`.

If you do not consider it part of this block then it should be moved to a separate story.

src/stories/Library/link-with-icon/link-with-icon.scss
```diff
@@ -0,0 +1,60 @@
...
+  margin-right: $s-md;
+  height: 40px;
+  width: 40px;
+  min-width: 40px;
+  min-height: 40px;
```
kasperg, line 51: Please consider if there is a reason behind using both `height`, `min-height`, `width` and `min-width` here.

Depending on your intentions I would expect one of them to be enough.

src/stories/Library/link-with-icon/link-with-icon.scss
```diff
@@ -0,0 +1,60 @@
...
+  border-radius: 50%;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+
+  &.rotate-180 {
+    transform: rotate(180deg);
```
kasperg, line 58: Please make this a modifier e.g. `link-with-icon__icon--rotate-180`.

src/stories/Library/link-with-icon/link-with-icon.scss
```diff
@@ -0,0 +1,60 @@
...
+  line-height: 160%;
+  width: 100%;
+  @include container(medium);
+
+  > svg {
+    display: block;
+    width: 35px;
+    margin-left: auto;
+    visibility: hidden;
+  }
```
kasperg, line 17: Please reconsider this.

As far as I can tell the intention is to target the `arrow` child element. In general we should not do so this way. In general: Without using classes or code comments it is hard to tell what is going on and why.

On the other hand I can see that [we seem to practice this in other situations](https://github.com/danskernesdigitalebibliotek/dpl-design-system/blob/develop/src/stories/Library/card-list-item/card-list-item.scss#L23) and you might have been inspired from them. If so then I would at least like us to use the mixin provided by `arrows.scss`.

src/stories/Library/link-with-icon/LinkWithIcon.tsx
```diff
@@ -0,0 +1,51 @@
...
+      )}
+    >
+      <div className={clsx("link-icon bg-identity-primary", linkIconClass)}>
+        <img className="invert" src={`icons/${folder}/${icon}.svg`} alt="" />
```
kasperg, line 45: Please also consider how you can move away from utility classes and towards BEM classes here.

src/stories/Library/link-with-icon/LinkWithIcon.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  const linkIconClass = clsx({
+    "rotate-180": linkType === "internal",
+  });
+
+  return (
+    <a
+      href={href}
+      target={target}
+      rel={target === "_blank" ? "noopener noreferrer" : undefined}
```
kasperg, line 39: Please consider if there is a reason for us to set `noopener` and/or `noreferrer`.

[According to MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel) `noreferrer` has the same effect as `noopener` so if we set `noreferrer`, `noopener` can be removed.

Also: Do we want to set either?

src/stories/Library/link-with-icon/LinkWithIcon.tsx
kasperg, line 45: :+1: for setting an empty `alt=""` as the icon is purely decorative.

src/stories/Library/link-with-icon/LinkWithIcon.stories.tsx
```diff
@@ -0,0 +1,63 @@
...
+    <LinkWithIcon
+      href="/search?q=storybook"
+      linkText="Search Link"
+      linkType="search"
+    />
+  </div>
+);
+
+export const LinkList = LinkWithIconListComponentTemplate.bind({});
+LinkList.decorators = [(Story) => <div>{Story()}</div>];
```
kasperg, line 63: Please reconsider this decorator.

As far as I can tell it adds a wrapper `<div>` which seems superfluous when you already have one added on L42.

Link with icon - Link List ⋅ Storybook 2023-12-03 14-37-20

src/stories/Library/link-with-icon/LinkTypes.ts
```diff
@@ -0,0 +1,7 @@
+export type LinkTypes = "internal" | "external" | "download" | "search";
+
+export type IconTypes =
```
kasperg, line 3: Please change the type to be in singular form.

In general we do not use plural for types which contain a single value which is the case here.

src/stories/Library/link-with-icon/LinkWithIcon.stories.tsx
```diff
@@ -0,0 +1,63 @@
...
+      },
+      defaultValue: "internal",
+    },
+  },
+  parameters: {
+    design: {
+      type: "figma",
+      url: "https://www.figma.com/file/Zx9GrkFA3l4ISvyZD2q0Qi/Designsystem?type=design&node-id=7477-39110&mode=design&t=d1dz4FVEPRORqxUN-4",
+    },
+    layout: "full",
```
kasperg, line 29: :+1: Thanks for remembering to add the design link to the story.

kasperg (changes requested):
:+1: Nice start.

There is some different things going on here:

1. The CSS needs to be adjusted to our BEM structure. If you need to read up on BEM [we have a few good links for reference in Zulip](https://reload.zulipchat.com/#narrow/stream/235451-dev-snak/topic/BEM/near/297829152).
2. The CSS needs to be adjusted accoridng to our decisions regarding more BEM and less utility classes.
3. I appreciate you doing a little bit of housekeeping as a part of this PR while keeping it in separate commits.

base.scss
```diff
@@ -11,6 +11,7 @@
 @import "./src/stories/Library/links/links";
 @import "./src/stories/Library/link-filters/link-filters";
 @import "./src/stories/Library/Arrows/arrows";
+@import "./src/styles/scss/container";
```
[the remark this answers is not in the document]
LasseStaus: Yeah this was an oversight.

src/stories/Library/link-with-icon/LinkWithIcon.stories.tsx
LasseStaus: Good catch, oversight from me

src/stories/Library/link-with-icon/LinkWithIcon.tsx
LasseStaus: I removed it entirely then. 

Noopener was indeed redundant, and i guess if we set and external link we are fine with sending the referer information along.

public/icons/README.md
```diff
@@ -0,0 +1,9 @@
+# Icon Usage Guidelines
+
+This folder contains SVG icons that are used in the design system. When using icons from this folder, please follow the guidelines below:
+
+- Icons located in public/ folder is the current source of truth. All icons will be placed in this folder. Ensure that icons placed gere are using the `class` attribute, and not `className` for SVG element. This is necessary for the icons to work properly when the folder is copied over using the "dapple"-setup.
```
kasperg, line 5:
```suggestion
- Icons located in public/ folder is the current source of truth. All icons will be placed in this folder. Ensure that icons placed gere are using the `class` attribute, and not `className` for SVG element. 
```

I do not think it should be relevant to mention Dapple at this level of documentation:

As I see it this is a general problem that just has not surfaced yet because 90% of all work (and very little) design has been going on in the CMS up until now.

public/icons/README.md
```diff
@@ -0,0 +1,9 @@
+# Icon Usage Guidelines
```
kasperg, line 1: :+1: for taking the time to write this. I suggest that we move it into the `/docs` folder to keep it a part of the general documentation.

If we need to then we can create a small README in this folder and then reference the larger file in `/docs`.

kasperg: In the future should expand the documentation to explain how developers should use the SVGs ie. inline the files when requiring animation using either copy/paste of techniques like `source()` in Twig and direct references with `` or `background-image` otherwise.

kasperg (approved):
:+1: Very nice update.

I especially appreciate you taking the time to improve the documentation.

This should be good to merge once the linting issues have been resolved. 

Please also remember to approve the changes in Chromatic if you are satisfied with them. I took a look and as I see it they look good.

PR #353, opened 2023-11-29
Add `EventPage` + `EventHeader`

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-14

#### Description
This pull request introduces two new components: `EventPage` and `EventHeader`. These components are used for the event-related pages of our application.

#### Screenshot of the result
image

image


src/stories/Library/event-header/EventHeader.tsx
```diff
@@ -0,0 +1,46 @@
...
+          size="large"
+          variant="filled"
+        />
+      </section>
+      <section className="event-header__visual">
+        <img
+          className="event-header__image"
+          src="/images/event_image.jpg"
+          alt=""
+        />
```
kasperg, line 36: Please remove this image from the repository.

We do not know anything about the copyright for this image. Consequently it should not be a part of the project. I suggest replacing it with an external url (potentially to the same image). This will also allow editors to test it out with images with different dimensions.

src/stories/Library/event-header/EventHeader.tsx
```diff
@@ -0,0 +1,46 @@
...
+        </div>
+        <Button
+          classNames="event-header__button"
+          label="Køb billet"
+          buttonType="none"
+          disabled={false}
+          collapsible={false}
+          size="large"
+          variant="filled"
+        />
```
kasperg, line 29: Please change this to a link.

While the element visually looks like a button in HTML semantics it is actually a link to an external page.

src/stories/Library/event-header/EventHeader.tsx
```diff
@@ -0,0 +1,46 @@
...
+      </section>
+      <section className="event-header__visual">
```
kasperg, line 31: I suggest replacing `<section>` with [`<figure>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figure).

There does not seem to be a clean cut solution on how to implement credits but [people on Stackoverflow has a nice discussion about the subject](https://stackoverflow.com/questions/57899136/how-to-implement-a-caption-and-photographer-credit-with-figure).

kasperg: Consider implementing a separate block for credited images.

In general this setup with an image with credits below seems to be used several places outside of the event header. I think it is OK not to do it here, now, but we may need to refactor it in the future.

src/stories/Library/event-header/EventHeader.tsx
```diff
@@ -0,0 +1,46 @@
...
+  date: string;
+};
+
+const EventHeader: FC<EventHeaderProps> = ({ title, date }) => {
+  return (
+    <header className="event-header">
+      <section className="event-header__content">
+        <div className="event-header__extra-padding">
+          <Tag hasBackground className="event-header__tag">
+            UDSTILLING
```
kasperg, line 16: Please do not make this all-caps in the reference.

All-caps is a visual representation that should be implementing in CSS.

src/stories/Library/event-header/EventHeader.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { Button } from "../Buttons/button/Button";
+
+type EventHeaderProps = {
+  title: string;
+  date: string;
+};
+
+const EventHeader: FC<EventHeaderProps> = ({ title, date }) => {
+  return (
+    <header className="event-header">
```
kasperg, line 12: :+1: for using `<header>` here. It makes sense.

src/stories/Library/event-header/EventHeader.tsx
```diff
@@ -0,0 +1,46 @@
...
+      <section className="event-header__content">
+        <div className="event-header__extra-padding">
```
kasperg, line 14: Please change the name of the `event-header__extra-padding` class.

It is a *very* specific name. If possible I would like to avoid the div entirely. I wondered if we could implement it like a three element grid (image, content and state) where all elements are in one column on mobile and in two colums on desktop. Image in one and content + state in the other.

If that is not possible then I would perfer a better class name.

src/stories/Library/event-header/EventHeader.tsx
```diff
@@ -0,0 +1,46 @@
...
+          collapsible={false}
+          size="large"
+          variant="filled"
+        />
+      </section>
+      <section className="event-header__visual">
+        <img
+          className="event-header__image"
+          src="/images/event_image.jpg"
+          alt=""
```
kasperg, line 35: Please do not specify an empty `alt` attribute here.

As I see it the image is not purely decoration. I think we should expect editors to provide proper alt text.

src/stories/Blocks/event-page/EventPage.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+type EventPageProps = {
+  title: string;
+  date: string;
+};
+
+const EventPage: FC<EventPageProps> = ({ title, date }) => {
+  return (
+    <article className="event-page">
+      <EventHeader title={title} date={date} />
```
kasperg, line 12: :+1: Very appropriate to start building an entire page for an event.

kasperg (changes requested):
:+1: I like the new changes.

The only thing I that needs to be addressed now is the usage of `<figure>`.

kasperg (dismissed):
:+1: LGTM

kasperg (approved):
:+1: LGTM

PR #362, opened 2023-12-05
Added flatpickr

JacobArrow:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-351

#### Description
We needed to find a find to share a datepicker across React and Drupal with the same styling applied both places.
This PR introduces flatpickr JS as the foundation for our new datepicker. The design system makes sure that flatpickr is styled the same way whereever it is used, and it is up to the respective app to implement flatpickr how it sees fit. The React app could for example use the Flatpickr React package and Drupal the native JS package. It does really not matte, because the flatpickr is the same. 

#### Screenshot of the result
The bare bone date calendar
image

The date calendar implemented as a datepicker
image

### DIFF
--- a/src/stories/Library/date-calendar/date-calendar.scss
+++ b/src/stories/Library/date-calendar/date-calendar.scss
@@ -1,0 +1,78 @@
+.flatpickr-calendar {
+  font-family: "Noto Sans JP";
+  font-style: normal;
+  font-weight: 500;
+  color: $c-text-primary-black;
+  padding: $s-md !important;
+  width: fit-content !important;
+
+  & * {
+    font-weight: 500 !important;
+  }
+
+  & .flatpickr-months {
+    & .flatpickr-month {
+      order: -1;
+      display: flex;
+      padding-left: 5px;
+
+      & .flatpickr-current-month {
+        display: flex;
+        align-items: center;
+        position: relative;
+        padding: 0;
+        left: 0;
+        width: auto;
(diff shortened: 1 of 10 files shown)


src/stories/Library/date-calendar/date-calendar.scss
```diff
@@ -0,0 +1,90 @@
...
+    & .flatpickr-prev-month,
+    & .flatpickr-next-month {
+      position: relative;
+      height: auto;
+      line-height: 100%;
+      display: flex;
+      align-items: center;
+
+      &:hover {
+        // inherited by flatpickr
```
spaceo, line 62: Great with comment!

src/stories/Library/date-calendar/date-calendar.scss
```diff
@@ -0,0 +1,90 @@
...
+    }
+  }
+  & .flatpickr-days {
+    & .flatpickr-day {
+      display: flex;
+      align-items: center;
+
+      &.prevMonthDay,
+      &.nextMonthDay {
+        color: rgba(57, 57, 57, 0.7);
```
spaceo, line 85: Could it be an idea, although the value is only used once, to make a local scss variable for this color? In that way we can describe the expected color, and it easier to decipher when looking at the code. Same goes for line 63.

spaceo (approved):
LGTM 👍  with one suggestion...

PR #368, opened 2023-12-06
Rich text / WYSIWYG styling. DDFFORM-5.

rasben:
This component is used for WYSIWYG content.
E.g. we cannot control classes of the underlying elements, so in
this case we will target HTML tags instead.

### DIFF
--- a/src/stories/Library/rich-text/RichText.tsx
+++ b/src/stories/Library/rich-text/RichText.tsx
@@ -1,0 +1,50 @@
+export interface RichTextProps {
+  text: string;
+}
+
+export const RichText = () => {
+  return (
+    <div className="rich-text">
+      <h2>Tad survive ensnare joy mistake courtesy Bagshot Row.</h2>
+      <p>
+        <a href="#">Ligulas step drops both?</a> You shall not pass! Tender
+        Tender respectable success. Valar impressive unfriendly bloom scraped?
+        Branch hey-diddle-diddle <strong>pony troublell sleeping</strong>.
+      </p>
+      <h3>North valor overflowing sort Iáve mister kingly money?</h3>
+      <p>
+        Curse you and all the halflings! Deserted anytime Lake-town burned caves
+        balls. Smoked lthilien forbids Thrain?
+      </p>
+      <ul>
+        <li>Adamant.</li>
+        <li>
+          Ligulas step drops <em>both? You shall not pass!</em>. Valar
+          impressive unfriendly bloom scraped? Branch hey-diddle-diddle pony
+          troublell sleeping during jump Narsil.
+        </li>
+        <li>Witch-king.</li>
+        <li>Precious.</li>
+        <li>Gaffers!</li>
+      </ul>
+      <h3>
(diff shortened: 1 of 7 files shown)


kasperbirch1 (approved):

src/stories/Library/rich-text/rich-text.scss
```diff
@@ -0,0 +1,84 @@
...
+  a {
+    text-decoration: underline solid black;
+    text-underline-offset: 0.3em;
+    transition: text-underline-offset 0.1s ease-in-out;
+
+    &:hover {
+      color: $color__text-primary-black;
+      text-underline-offset: 0.2em;
+    }
+  }
```
kasperg, line 19: Consider whether this should include the styling from `.link-tag`.

If there is a reason we do not do so then I would appreciate a comment mentioning why.

src/stories/Library/rich-text/rich-text.scss
```diff
@@ -0,0 +1,84 @@
...
+      text-align: center;
+    }
+
+    @include media-query__medium {
+      padding-left: $_dash-width--wide + $s-lg;
+
+      &::before {
+        width: $_dash-width--wide;
+      }
+    }
```
kasperg, line 56: :+1: Nice use of variables.

src/stories/Library/rich-text/rich-text.scss
```diff
@@ -0,0 +1,84 @@
+// This component is used for WYSIWYG content.
+// E.g. we cannot control classes of the underlying elements, so in this case
+// we will target HTML tags instead.
+.rich-text {
```
kasperg, line 4: :+1: Excellent commenting.

kasperg (approved):
:+1: LGTM. I have a few comments for you to consider before you merge.

src/stories/Library/rich-text/RichText.tsx
```diff
@@ -0,0 +1,52 @@
+export interface RichTextProps {
+  text: string;
+}
+
+export const RichText = () => {
```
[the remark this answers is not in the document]
rasben: @kasperbirch1 because i dont know any better :D 
isnt this how you guys usually do it in storybook?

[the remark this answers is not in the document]
kasperbirch1: I have noticed that this is how it is done in some places in the code. but I'm pretty sure it's unnecessary. As I see it, it should only be exported as default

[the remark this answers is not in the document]
rasben: Fair enough, I'll fix it :)

src/stories/Library/rich-text/rich-text.scss
rasben: Link tag is a different thing - it's a link with an arrow and stuff, and overwriting font-size etc.
This is just a usual inline-link.

But, you're right. This is useful in other cases too.
I'll move this chunk of code into a placeholder of it's own.

PR #370, opened 2023-12-06
Handling releases with versioned assets

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-328

#### Description

Splitting deployment into two workflows
One for handling releases invoked by tagging.
Another for PR changes.

The original idea was to to make the pr create a tag which the other workflow would react upon. But that flow is not directly supported.

The bundle script has been expanded to support writing a version file.


.github/workflows/create-release-on-pr-changes.yml
```diff
@@ -0,0 +1,91 @@
...
+
+      - name: Delete existing releases
+        uses: dev-drprasad/delete-tag-and-release@v1.0
+        with:
+          tag_name: ${{ env.TAG_NAME }}
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+
+
+      - name: Create or update tag
+        uses: EndBug/latest-tag@latest
```
kasperg, line 57: Please use an actual version here instead of `latest` - at least `v1`.

Chasing a moving target without the limitations of semantic versioning or the like is seldomly a good idea.

.github/workflows/create-release-on-pr-changes.yml
```diff
@@ -0,0 +1,91 @@
...
+      - name: Adding summary
+        run: |
+          echo "Release created 🚀😎 at: ${{ steps.create-release.outputs.url }}" >> $GITHUB_STEP_SUMMARY
+
+      - name: Release NPM package
+        # Version 0.0.0-SHA is a schema that supports semantic versioning but
+        # should sink below proper versions.
+        # Output package.json to provide insight and help debugging
+        # Using branch names as tags allows other projects to track unreleased
+        # development.
```
kasperg, line 80: Please move comments about version names closer to where `NPM_VERSION` is defined.

At this point it is just a variable.

bundle.sh
```diff
@@ -6,4 +6,9 @@ cp -r ./public/icons ./build/icons
 cp -r ./src/styles/css ./build/css
 cp -r ./src/styles/fonts ./build/fonts
 find ./src -name "*.js" -exec cp {} ./build/js \;
+
+if [ -n "$VERSION" ]; then
+   echo $VERSION > ./build/version.txt
+fi
```
kasperg, line 12: :+1: Nice!

.github/workflows/create-release-on-pr-changes.yml
```diff
@@ -0,0 +1,91 @@
...
+        id: resolve-release-vars
+        run: |
+          RELEASE_BRANCH=${{ github.event.pull_request.head.ref }}
+          RELEASE_SHA=${{ github.event.pull_request.head.sha }}
+          TAG_NAME=$(echo "branch-${RELEASE_BRANCH}")
+          echo "RELEASE_BRANCH=$RELEASE_BRANCH" >> "$GITHUB_ENV"
+          echo "RELEASE_SHA=$RELEASE_SHA" >> "$GITHUB_ENV"
+          echo "TAG_NAME=$TAG_NAME" >> "$GITHUB_ENV"
+          echo "NPM_VERSION=0.0.0-${TAG_NAME}-${RELEASE_SHA}" >> "$GITHUB_ENV"
+          echo "DIST_FILENAME=dist-$RELEASE_BRANCH-$RELEASE_SHA.zip" >> "$GITHUB_ENV"
```
kasperg, line 21: The more I think about it the more I do not agree with adding the SHA to the file name for branches.

As I see it, because of Docker caching we need to make an update to composer to get the latest build of a branch. This will still be the case when adding the SHA to file names - but it will be very obvious because builds will be failing if you do not do so.

On the other hand having a development environment or release branch with a continually updated version will be practically impossible because the SHA for the design system will always be changing. You will have to commit an update the referenced file each time. One developers ability to reinstall dependencies will break because another developer has merged a PR to the design system removing the existing file.

I understand the wish to use a fixed version of these projects but now that we have proper builds on tags I suggest that we use this instead. When a developer wishes to use a static version of the design system - for a demo environment or a proper release - they create a tag. e.g. 2023.50.0-rc1 and reference the files uploaded to that build.

.github/workflows/create-release-on-pr-changes.yml
```diff
@@ -0,0 +1,91 @@
+name: Create release on PR changes
+
+on:
+  pull_request:
```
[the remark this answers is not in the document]
kasperg: Please remember to exclude work by Dependabot.

.github/workflows/create-release-on-pr-changes.yml
spaceo: 👍

.github/workflows/create-release-on-pr-changes.yml
spaceo: Since I have introduced docker [build cache busting](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/534) it should not be necessary anymore.

.github/workflows/create-release-on-pr-changes.yml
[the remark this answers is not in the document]
spaceo: Done

kasperg (approved):
:+1: Let's roll!

PR #371, opened 2023-12-07
Fix/pay-button

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-334

#### Description

This PR makes sure that any text inside a button is never with text decorations, even if there is an <a> tag inside it. We never want an underline, and instead of using the utility class "hide-linkstyle" I am moving this responsibility to the main button class.

#### Screenshot of the result
n/a

Sibling to [this PR](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/754).


JacobArrow (approved):
Neat! I like it

Adamik10:
Follow in #384

PR #389, opened 2023-12-14
DDFFORM 59 arrangement liste

LasseStaus:
- Update 'Noto sans JP' to 'Noto sans'
- Added EventListItem for displaying an event.
- Added EventList as a block, in order to display multiple events.

#### Link to issue

[jira issue](https://reload.atlassian.net/jira/software/c/projects/DDFFORM/boards/485?selectedIssue=DDFFORM-59)

### Changes
- Replaced 'Noto Sans JP' with 'Noto Sans' in the design system.
- Implemented EventListItem component for individual event presentation.
- Created EventList block for displaying multiple EventListItem components.

#### Description

I've added a styling for a single event and a list containing multiple events. 

See jira issue for an overview of what is included. 

I did however include a "category-tag", since this has been done elsewhere too i thought i would do it aswell. 

I also added styling for an image since the media library is almost ready. 

Go commit by commit. See comments underneath the screenshots and in commit messages.

#### Screenshot of the result
image
image
image

I've messed around a bit in this issue with using 'normal' grid, grid-areas, flex box or a combination. Please comment on your thoughts for the css layout. 

Test the component out with the controls to see if it behaves as expected. 

- Are texts' properly truncated 
- If certain items are blank, does the content move, etc. 

The "date" was not a part of the figma mobile layout, i assume this is a mistake and have added it for mobile. I placed on the right side on mobile, comment on whether you agree. 

There are 270~ ui changes due to the font change. 

Lotte / Nino has approved of the "style" changes that accompany the switch of the font.


rasben:
@LasseStaus Before this can be reviewed, you need to take into consideration that there's 275 UI tests that are changed.

It appears to be because you've updated the font, and maybe it just needs to be approved en-masse

src/stories/Blocks/event-list/EventList.stories.tsx
```diff
@@ -0,0 +1,11 @@
+import { ComponentStory, ComponentMeta } from "@storybook/react";
+import EventList from "./EventList";
+
+export default {
+  title: "Blocks / Event List",
+  component: EventList,
+} as ComponentMeta<typeof EventList>;
```
kasperbirch1, line 7: Please add Figma link to design.

Example:
```
import { withDesign } from "storybook-addon-designs";

  parameters: {
    design: {
      type: "figma",
      url: "https://www.figma.com/file/Zx9GrkFA3l4ISvyZD2q0Qi/Designsystem?type=design&node-id=7477%3A39048&mode=dev",
    },
  ```

File: src/stories/Blocks/event-list/EventListData.ts [hunk 2]
```diff
@@ -0,0 +1,50 @@
+import { EventListItemProps } from "../../Library/event-list-item/EventListItem";
```
kasperbirch1, line 1: Would it make sense to have this as part of the arguments in the story?

src/stories/Blocks/event-list/event-list.scss
```diff
@@ -0,0 +1,9 @@
+.event-list {
```
kasperbirch1, line 1: Please use the new ` @include layout-container`

src/stories/Library/event-list-item/event-list-item.scss
```diff
@@ -0,0 +1,152 @@
...
+  text-decoration: none;
+  display: grid;
+  padding: $s-lg;
+  row-gap: $s-md;
+  max-width: $layout__max-width;
+  background-color: #fcfcfc;
+  margin: 0 auto;
+}
+.event-list-item__image-container {
+  background: green; // Placeholder for image background / component
```
kasperbirch1, line 13: Can you use the `@extend %identity-placeholder;` here?

kasperbirch1 (approved):

src/stories/Library/event-list-item/event-list-item.scss
LasseStaus: I couldn't use this directly no. But i have added a version of it that is using ::before pseudo that i am now using. Can be seen in the next commits.

src/stories/Library/event-list-item/EventListItem.tsx
```diff
@@ -0,0 +1,49 @@
...
+  location,
+  price,
+  href,
+}) => {
+  return (
+    <a href={href} className="event-list-item arrow__hover--right-small">
+      <div className="event-list-item__image-container">
+        <img src={image} alt={title} className="event-list-item__image" />
+      </div>
+      <div className="event-list-item__content">
```
kasperg, line 31: I wonder if it is possible to drop the `event-list-item__image-container` wrapper.

Perhaps it is just my personal dislike of unnecessary markup :smile:.

src/stories/Library/event-list-item/EventListItem.tsx
```diff
@@ -0,0 +1,49 @@
...
+        <Tag hasBackground className="event-list-item__tag">
+          foredrag
+        </Tag>
```
kasperg, line 34: Please make the tag configurable as well.

I cannot tell why it should not be.

src/stories/Library/event-list-item/EventListItem.tsx
```diff
@@ -0,0 +1,49 @@
...
+        <p className="event-list-item__date">{date}</p>
+        <h2 className="event-list-item__title">{title}</h2>
+        <p className="event-list-item__description">{description}</p>
+        <div className="event-list-item__location-wrapper">
+          <p className="event-list-item__location">{location}</p>
+        </div>
+        <div className="event-list-item__schedule">
+          <time className="event-list-item__time">{time}</time>
```
kasperg, line 42: Consider whether it is appropriate to add a [`datetime` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time) for the `<time>` element - and whether this or the date is the right element to do it.

I know that adding the correct value corresponding to the displayed will be difficult but with a comment it might be OK to hard code the value to show that it is important to actually have it set.

src/stories/Library/event-list-item/event-list-item.scss
```diff
@@ -0,0 +1,176 @@
...
+  margin: 0 auto;
+  grid-template-areas:
+    "image"
+    "content";
+  grid-template-rows: auto;
+  row-gap: $s-md;
+  max-width: $layout__max-width;
+
+  > svg {
+    display: none;
```
kasperg, line 16: Consider if this is necessary.

The `show-svg-on-hover` mixin already has ` > svg { visibility: hidden }` by default. If that is not enough here then please add a comment explaining why.

If this is a general problem then I think we should address it if the mixin if possible.

src/stories/Library/event-list-item/event-list-item.scss
```diff
@@ -0,0 +1,176 @@
+.event-list-item {
+  @include show-svg-on-hover();
+  text-decoration: none;
+  display: grid;
+  padding: $s-lg;
+  background-color: #fcfcfc;
```
kasperg, line 6: Please consider if this value is set right.

As far as I can tell in [Figma the background color of event-list-items`](https://www.figma.com/file/Zx9GrkFA3l4ISvyZD2q0Qi/Designsystem?type=design&node-id=7477-45287&mode=design&t=Tyjdibqi0YgvStKK-0) is white.

Also: In general we should colors from `variables.colors.scss`. If the color is not already available we should add it.

src/stories/Library/event-list-item/event-list-item.scss
```diff
@@ -0,0 +1,176 @@
...
+  @include typography($typo__h4);
+  @extend %text-truncate;
+  margin-bottom: $s-sm;
+}
+
+.event-list-item__description {
+  grid-area: description;
+  @include typography($typo__small-caption);
+  @extend %text-ellipsis-two-lines;
+  -webkit-line-clamp: 3;
```
kasperg, line 74: Consider if we can make `text-ellipsis-two-lines` into a mixin with a parameter.

It seems as if we would like to add a configurable clamp here.

src/stories/Library/event-list-item/event-list-item.scss
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  .event-list-item__title {
+    margin-bottom: unset;
+    grid-area: title;
+  }
+
+  .event-list-item__description {
+    margin-bottom: unset;
+    -webkit-line-clamp: 2;
```
kasperg, line 141: Consider if this should use `%text-ellipsis-two-lines`.

If we want to use line clamp this seems to be the way we do it.

src/styles/scss/tools/variables.fonts.scss
```diff
@@ -1,44 +1,41 @@
...
-$font__body: "Noto Sans JP", sans-serif;
+$font__body: "Noto Sans", sans-serif;
 $font__accent: "Gochi Hand", cursive;
 $font__weight--normal: 400;
 $font__weight--semi-bold: 500;
 $font__weight--bold: 700;
 
-// Since we host the fonts ourselves, they may need to be updated in the future
-// We use this online open source tool to retrieve fonts & their css definition:
-// https://gwfh.mranftl.com/fonts
```
kasperg, line 10: Is there a reason you dropped this comment?

If you used this tool to generate the new font definitions then I think it is very appropriate to keep around.

src/styles/fonts/notosans-bold-webfont.woff
kasperg, line 1: With this change [preloading will no longer work](https://github.com/danskernesdigitalebibliotek/dpl-cms/blob/f9d385c4e2fe33ac24180bb0a446f5f9f2a3daac/web/themes/custom/novel/templates/layout/html.html.twig#L36-L50) as the file names have changed.

Please update the paths in the CMS accordingly.

kasperg: In general I think this is a nice change to a somewhat odd decision in the initial design.

I would appreciate a more elaborate description in the commit comment of what issues we were encountering. This might be important for future reference if for instance designers ask why the font suddenly looks different.

kasperg (changes requested):
:+1: Well done. I think this demonstrates a nice implementation or our BEM structure and sensible HTML.

I have some comments for you to consider before this can be merged.

src/stories/Library/event-list-item/event-list-item.scss
LasseStaus: Added a mixin for this with a default of 2, and the ability to change lines as parameter. Refactored existing usages and removed the placeholder.

src/stories/Library/event-list-item/event-list-item.scss
LasseStaus: Added a mixin for this with a default of 2, and the ability to change lines as parameter. Refactored existing usages and removed the placeholder.

src/styles/scss/tools/variables.fonts.scss
LasseStaus: I did not use what was commented before to generate the new fonts. I used [font squirrel](https://www.fontsquirrel.com/) this time, but there are other generators. 

I don't think there is a particular reason to use their old generator, but that every generator is fine. As such i feel the comment doesn't provide anything, but if you want i do not mind adding a comment explaining these are generated with fontsquirrel etc. like the old comment.

src/stories/Library/event-list-item/event-list-item.scss
LasseStaus: Apparently there never was a color for setting black white and grey as background-color: in the system. 

Background-colors using variables, were being set using the text variable. I've added black, white and grey global colors and refactored this. 

And yeah this should have be set to white from the start 🥳

src/stories/Library/event-list-item/EventListItem.tsx
LasseStaus: Hmm i am not sure. 

I want to keep the wrapper, so i can put the css for the background on that element, and if there is no image not need to render any additional element instead of it. 

I am genereally unsure what is the best approach to handle image positioning, also considering we are gonna be using focal point etc. 

I felt this approach, with the css provided to both image-container and the __image itself, would create a layout that would position the image correctly, give the "grid"-structure needed, and have a fallback to the image placeholder with colors incase the image was not there, or something happened in while loading.

src/stories/Library/event-list-item/event-list-item.scss
LasseStaus: The element would still be rendered in mobile even tho the visibility was hidden, and i didn't want that, but I see that the code for ```show-svg-on-hover``  has been changed. 

I did another ticket, before the refactor of arrows, and now i just did it the same way. 

but i can see that benja has added
``` @include media-query__small("max-width") {
      display: none;
    }```
    to the
     
    ```show-svg-on-hover``` mixin. 
    
    which makes my code double confetti. 
    
    it has been removed! 🥳

File: src/styles/fonts/notosans-bold-webfont.woff
Comment (LasseStaus) [reply src/styles/fonts/notosans-bold-webfont.woff:1]: I have updated this in a commit for the CMS

File: src/styles/scss/tools/variables.fonts.scss (hunk 12 above)
Comment (kasperg) [reply src/styles/scss/tools/variables.fonts.scss:10]: I think there is value in providing information about the font generator used since this in the end can effect how the font is rendered to browsers.

- Google Web Fonts Helper includes version information about fonts in the generator and in the file name. This makes it possible to check if we are using a different versions and potentially track changes to a changelog.
- Font Squirrel has a lot of customization options. For future reference it is important to know what options the current webfonts were generated with - basic, optimal or expert. If expert even more information may be required :smile:.

All in all font generation can be a somewhat opaque process. If we are updating fonts and seeing unexpected changes it is important to be able to limit the factors which may cause these.

**Review by kasperg — APPROVED** [review]
:+1: Thank you for making the effort implementing these changes.

File: src/stories/Library/event-list-item/event-list-item.scss [hunk 13]
```diff
@@ -12,8 +12,12 @@
   row-gap: $s-md;
   max-width: $layout__max-width;
 
-  > svg {
-    display: none;
+  // Extending display none, as mixin show-svg-on-hover
+  // is only working up to __small breakpoint
+  @include media-query__medium("max-width") {
+    > svg {
+      display: none;
```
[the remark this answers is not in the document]
LasseStaus: I will write this down and do it as part of the layout css adjustments coming in shortly instead of this pr. This way i can also refactor existing use cases.

PR #390, opened 2023-12-15
Fix rich text + Re-factor layout max width as a mixin

rasben:
[Make sure that rich-text has spacing and margin. DDFFORM-5](https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/390/commits/dd485fb3e1a484e554a62a8195b8f4affa8e5f40) 

It worked in Storybook, but not in Drupal.
That is because Storybook adds its own wrapper, which made it appear
correct.

----

Add a layout-container mixin, and replace any existing max-widths.

### DIFF
--- a/src/stories/Library/rich-text/rich-text.scss
+++ b/src/stories/Library/rich-text/rich-text.scss
@@ -5,10 +5,7 @@
   @include typography($typo__rich-text-body);
 
   // Setting a max-width to increase readability for sentences.
-  max-width: $layout__max-width--text;
-  margin: auto;
-  padding-left: $layout__page-padding;
-  padding-right: $layout__page-padding;
+  @include layout-container($layout__max-width--text);
 
   a {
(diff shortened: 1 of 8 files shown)


src/stories/Library/rich-text/rich-text.scss
```diff
@@ -6,6 +6,9 @@
 
   // Setting a max-width to increase readability for sentences.
   max-width: $layout__max-width--text;
+  margin: auto;
```
kasperbirch1, line 9: shouldn't this be made as a mixin or a placeholder?

kasperbirch1 (approved):

kasperbirch1 (commented):
x

kasperbirch1 (changes requested):
x

kasperbirch1 (approved):

PR #395, opened 2023-12-18
fix old non responsive stories

LasseStaus:
- Updated mobile responsiveness for modalCTA item component.

#### Link to issue

I found this as a part of going through UI tests as part of changing the fonts. 

#### Description

Found a component that was not responsive. Made it responsive. 

#### Screenshot of the result

image

See the UI test that must be accepted in order to see the change.

### DIFF
--- a/src/stories/Library/Modals/modal-cta/modal-cta.scss
+++ b/src/stories/Library/Modals/modal-cta/modal-cta.scss
@@ -9,12 +9,22 @@
 }
 
 .modal-cta__buttons {
-  display: grid;
-  justify-content: center;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
 }
 
 .modal-cta__link {
   display: flex;
   justify-content: center;
   margin-top: $s-md;
+}
+
+.modal-cta__button {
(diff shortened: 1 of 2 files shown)


src/stories/Library/Modals/modal-cta/modal-cta.scss
```diff
@@ -9,12 +9,22 @@
 }
 
 .modal-cta__buttons {
-  display: grid;
-  justify-content: center;
+  display: flex;
```
kasperbirch1, line 12: I don't understand this part of the change. Shouldn't it give the same result as:
```
display: grid;
justify-content: center;
```

LasseStaus: before, the button had ':0 96px' padding. In this case what you are saying would work. 

Since there is no padding on mobile now, justify-content would squeeze it together. 

What i could've done i think, is use justify-content: stretch instead and that would've done the same thing. 

Now i just used flex instead, but probably would've been faster to just use justify-content: stretch

PR #397, opened 2023-12-18
Fix overflow on x-axis

JacobArrow:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-183

#### Description
This fixes an issue where the footer will take up too much space on the x-axis and overflow the main content. When using window sizing like vw or vh, scrollbars are counted in, and this will add additional unneeded space.


spaceo (approved):
LGTM 👍

PR #399, opened 2023-12-19
Add free-styled styling for WYSIWYG tables. DDFFORM-98

rasben:
#### Link to issue

The WYSIWYG (rich-text) in Drupal supported tables, but I had forgotten to make the styling in the design system.
There is no table design in figma, so i've just winged something, based on what I had laying around from another project.

#### Description

https://reload.atlassian.net/browse/DDFFORM-98


kasperbirch1 (approved):
LGTM 👍

PR #402, opened 2023-12-19
Layout units - horizontal and vertical.

rasben:
After a discussion today, we agreed to introduce layout unit components. These units are used for centering stuff, and adding padding, and for adding spacing between individual elements.
The idea is that this is done using markup wrappers, but I've also made them so they can be used by individual components using placeholders.

### DIFF
--- a/base.scss
+++ b/base.scss
@@ -1,6 +1,7 @@
 @import "./src/styles/scss/reset";
 @import "./src/styles/scss/skeleton";
 @import "./src/styles/scss/tools";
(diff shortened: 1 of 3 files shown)


kasperbirch1:
I don't have much experience with "layout" CSS classes, so it's difficult for me to determine what works best. 

@LasseStaus is there a .layout__horizontal-spacing missing in your commit, or is it because it's an example? 

I think Lasse's naming is the best. But let's talk about it tomorrow.

PR #405, opened 2023-12-20
Only show scrollbar when content is overflowing

JacobArrow:
overflow: scroll, will always show a scrollbar, which is very ugly


Dresse (approved):
LGTM 👍

PR #408, opened 2023-12-21
Added explicit focus  styles for buttons

JacobArrow:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-55

#### Description
To comply with the laws of WCAG, we need a distinctive style for focused buttons.
This needs to be an outline with at least 2px wideness, and a clear difference in contrast

#### Screenshot of the result
button focus


spaceo (approved):
LGTM 👍

PR #409, opened 2023-12-27
Padded layout in Storybook

JacobArrow:
#### Description
By overriding the default styling in storybook, we do not allow the padded layout, which was the default setting.
By overriding the default layout with "fullscreen", we can set a padded layout where needed.


Adamik10 (approved):
Get those internal classes out of my bus! Wroom wroom 🚌

PR #410, opened 2023-12-27
Focus outline on dropdowns

JacobArrow:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-216

#### Description
To comply with the laws of WCAG, we need a distinctive style for focused dropdown/select boxes.
This needs to be an outline with at least 2px wideness, and a clear difference in contrast

#### Pictures
Dropdown default
image

Dropdown focused
image

Accordion default
image

Accordion focused
image


Adamik10 (approved):
Very very very nice 💯

PR #411, opened 2023-12-27
Updated warning color to a slightly darker red

JacobArrow:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-235

#### Description
To accommodate the contrast laws of  WCAG, we need to use a warning color with higher contrast. By adjusting the current warning color to a slightly darker red, we can accomplish that.

#### Screenshot of the result

Before
image

After
image


Adamik10 (approved):
👍 👍 👍

PR #420, opened 2024-01-04
Card slider, and placeholder/pagefold refactor. DDFFORM-86

rasben:
Remove bem-namics rule from stylelint.

    It appears that it does not work as intended:
    It reports errors where it shouldnt, and doesnt
    report errors where it should.

    Ontop of that, it seems to cause a bug with // stylelint-disable-line.

    A lot of the files we choose to ignore it anyway,
    so we might aswell just remove it.

----

    Card slider, and placeholder/pagefold refactor. DDFFORM-86

    - Introduce a 'card' story, which corresponds to the 'news card'
      in Figma.
    - Cards can support images, or just a teaser text, shown in the
      page fold, placeholder style.
    - I've simplified the placeholder/pagefold functionality,
      using %placeholders and making them responsive, instead of
      defining specific sized HTML elements.
    - I've also included the 'SwiperJS' library, for the slider.
    - SwiperJS was chosen, as the elements in the slider needs to
      be able to be different sizes, whilst also allowing 'peeking',
      to the next slide.

-----

    Introduce more card view modes. DDFFORM-87

    Also re-factors the truncate placeholder, to be a
    mixin instead with the line amount as a parameter.

### DIFF
--- a/src/stories/Library/card/card.scss
+++ b/src/stories/Library/card/card.scss
@@ -1,0 +1,108 @@
+.card {
+  $_card-mobile-peek: 50px;
+
+  position: relative;
+  width: 200px;
+  max-width: 100%;
+  word-break: break-word;
+
+  .slider & {
+    // We dont want the max-width to be 100%, as we want peek-effect on mobile.
+    max-width: calc(100vw - #{$_card-mobile-peek});
+  }
+
+  @include media-query__name($breakpoint__large-card) {
+    width: 440px;
+  }
+
+  img {
+    height: 100%;
+    width: 100%;
+    object-fit: cover;
+    object-position: center;
+  }
+
+  a {
+    text-decoration: none;
+  }
+}
+
+.card__media {
+  aspect-ratio: 1/1;
+  box-sizing: border-box;
+  background-color: var(--tint-color-120);
+  margin-bottom: $s-sm;
+}
+
+.card--has-image {
+  // This is necessary to get object-fit to work.
+  * {
+    height: 100%;
+  }
+}
+
+.card--has-no-image {
+  @extend %identity-pagefold;
+
+  .card__media {
+    @extend %identity-placeholder;
+  }
+}
+
+.card__placeholder-text {
+  @include typography($typo__card-placeholder);
+  @extend %identity-placeholder-inner-padding;
+  display: flex;
+  height: 100%;
+  overflow: hidden;
+
+  align-items: end;
+  color: $color__text-primary-white;
+}
+
+.card[data-variant="large"] {
+  width: 327px;
+
+  @include media-query__name($breakpoint__large-card) {
+    width: 551px;
+  }
+
+  .card__media {
+    // Odd aspect ratio, taken from the design.
+    aspect-ratio: 551/312;
+  }
+}
+
+.card__tags {
+  display: flex;
+  margin-bottom: $s-sm;
+}
+
+.card__tag {
+  @include typography($typo__byline-tags);
+}
+
+// Placing spacing and spacer line between tags.
+.card__tag + .card__tag {
+  position: relative;
+  margin-left: $s-sm;
+  padding-left: $s-sm;
+
+  &::before {
+    content: "";
+    position: absolute;
+    top: $s-xs;
+    bottom: $s-xs;
+    left: 0;
+    width: 1px;
+    background-color: $color__text-primary-black;
+  }
+}
+
+.card__tag--type {
+  color: $color__text-inactive;
+}
+
+.card__title {
+  @include typography($typo__h3);
+}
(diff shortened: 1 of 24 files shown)


src/styles/scss/tools/variables.colors.scss
```diff
@@ -13,9 +13,15 @@ $color__identity-primary: hsl(
...
 );
 
+$color__page-bg: $color__global-primary;
+
 $color__text-primary-white: $color__global-white;
 $color__text-primary-black: $color__global-black;
 $color__text-secondary-gray: $color__global-grey;
+$color__text-primary-white: #fff;
+$color__text-primary-black: #000;
+$color__text-secondary-gray: #484848;
```
LasseStaus, line 23: I think all are these are duplications to the ones just above aren't they?

in ddfform-59 i added global black/white/grey and changed the __text ones to use those.

src/styles/scss/tools/placeholder.scss
```diff
@@ -1,4 +1,3 @@
-// stylelint-disable plugin/stylelint-bem-namics
 %text-truncate {
   white-space: nowrap;
   overflow: hidden;
```
LasseStaus, line 3: Can we remove this entirely, and just use the mixin instead? 

I see you added it for with param 1 line for .card__tags, and now i think it's only used one place in event-list. 

Up to you.

src/stories/Library/slider/Slider.stories.tsx
```diff
@@ -0,0 +1,72 @@
...
+    title="Bøger som har gjort en forskel for romanens udvikling"
+    typeTag="Nyhed"
+    dateTag="06 Dec 2022"
+    image={image}
+  />
+);
+
+const cardNoImage = (
+  <Card
+    title="Fars legestue, hver onsdag"
```
LasseStaus, line 47: I see what you did there 😆

src/stories/Library/slider/Slider.tsx
```diff
@@ -0,0 +1,75 @@
+import { FC, ReactNode, useEffect } from "react";
+import { Helmet } from "react-helmet";
+
+type SliderProps = {
+  title?: string;
+  items: ReactNode[];
+};
+
+const Slider: FC<SliderProps> = ({ title, items }) => {
```
LasseStaus, line 9: Just commenting it here. 

This could be storybook specific, havenøt testet in drupal yet. 

The slider seems to only be browsable with clicking on the arrow, i guess it should be scrollable too for better ux?

src/stories/Library/slider/slider.scss
```diff
@@ -0,0 +1,75 @@
...
+.slider__control {
+  width: 115px;
+  max-width: 100%;
+  margin-left: auto;
+  transition: opacity 0.2s ease-in-out;
+  cursor: pointer;
+
+  &[disabled] {
+    opacity: 0;
+    width: 0;
```
LasseStaus, line 39:
```suggestion

```
When getting to the end of either direction in the slider, the oppising arrow will "jump" before turning invisible when i test. 

I think this can be changed by just removing this width css.

src/stories/Library/card/card.scss
```diff
@@ -106,3 +63,102 @@
...
+
+    .card__tag {
+      @include typography($typo__byline-tags--small);
+    }
+
+    .card__title {
+      @include typography($typo__h4);
+    }
+  }
+}
```
LasseStaus, line 154: I dont fully understand the implications of the combined css, or if i'm hitting the all the correct parts. Possibly it works just as intended. 

So i'll just highlight that a small card with no image, has no text inside it the identityplaceholder when above the $breakpoint__large-card, which i understand to be the medium breakpoint of 1024px. 

If this is intended then i guess all is well, but it seemed strange to me that the text for would disappear on larger screens for seemingly no reason.

LasseStaus (approved):
Nice slider! 

The slider "arrow" is jumping when i test it, i think that could be changed. 

Otherwise i added some comments you can consider. 

I think potentially there could be done something with how far to the left the slider positions itself, depending on screen size. On large screens it would go far out to the side which might not be desired, but i think we should just take that at the demo. 

I have approved so you don't need to wait for that to merge, but i think at least the jumping arrow should be addressed.

src/stories/Library/slider/Slider.tsx
rasben: This is something I'd like to talk with DDF about during a demo - for now, i'll keep it like this.

src/styles/scss/tools/variables.colors.scss
rasben: You're right, it's a merge conflict, i'll fix it.

src/stories/Library/card/card.scss
rasben: This is correct and intended
On mobile, the "small" card is the same size as the x-large, so there is actually space for the placeholder text, as it also becomes smaller.
On desktop, the placeholder text will become bigger, but the small box doesnt, so there's no space.

This is also something I want to take with DDF during demo.

PR #427, opened 2024-01-08
Implement `show more` functionality as a utility

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-130

#### Description
This pull request introduces the `show more` feature as a utility for use in Drupal / dpl-cms, specifically designed for `HorizontalTermLine` and `Tags`. 

It also includes a refactor of the `EventHeader` component, enabling it to manage multiple tags if needed.

#### Screenshot of the result

### DIFF
--- a/src/stories/utils/show-more.js
+++ b/src/stories/utils/show-more.js
@@ -1,0 +1,36 @@
+document.addEventListener("DOMContentLoaded", () => {
+  const categories = document.querySelectorAll("[data-show-more-item]");
+  const toggleButton = document.querySelector("[data-show-more-button]");
+  const initialVisibleItems =
+    document.querySelector("[data-show-more-initial-visible-items]") ?? 2; // default to 2 if not set
+
+  // if there are no categories or only one category, hide the toggle button
+  // or if there are less categories than the initial visible items, hide the toggle button
+  if (categories.length <= 1 || categories.length <= initialVisibleItems) {
+    toggleButton.style.display = "none";
+    return;
+  }
+
+  // Get the show more and show less text from data attributes or use defaults
+  const showMoreText =
+    toggleButton.getAttribute("data-show-more-text") ?? "Show more";
+  const hideMoreText =
+    toggleButton.getAttribute("data-show-more-hide-text") ?? "Show less";
+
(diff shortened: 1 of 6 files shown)


src/stories/utils/show-more.js
```diff
@@ -0,0 +1,36 @@
+document.addEventListener("DOMContentLoaded", () => {
+  const categories = document.querySelectorAll("[data-show-more-item]");
+  const toggleButton = document.querySelector("[data-show-more-button]");
+  const initialVisibleItems =
+    document.querySelector("[data-show-more-initial-visible-items]") ?? 2; // default to 2 if not set
+
+  // if there are no categories or only one category, hide the toggle button
+  // or if there are less categories than the initial visible items, hide the toggle button
+  if (categories.length <= 1 || categories.length <= initialVisibleItems) {
+    toggleButton.style.display = "none";
```
rasben, line 10: I'd really prefer you use a class instead, e.g. `.is-hidden` or something.
I dont really want inline styles when we can avoid it.

src/stories/utils/show-more.js
```diff
@@ -0,0 +1,36 @@
...
+    return;
+  }
+
+  // Get the show more and show less text from data attributes or use defaults
+  const showMoreText =
+    toggleButton.getAttribute("data-show-more-text") ?? "Show more";
```
rasben, line 16: Perhaps, the fallback should be something that is non-text, so it works even without translation - maybe something as simple as `+` and `x` ?

rasben: or emoji, whatever

src/stories/utils/show-more.js
```diff
@@ -0,0 +1,36 @@
...
+  const hideMoreText =
+    toggleButton.getAttribute("data-show-more-hide-text") ?? "Show less";
```
rasben, line 18: same comment ^

rasben (changes requested):
Overall, looks good :) A few comments.
Instead of Javascript, you could also have done it with `<summary>` and `<details>`, but i respect that your solution might be easier to maintain.

https://www.w3schools.com/tags/tag_summary.asp

rasben (approved):

PR #431, opened 2024-01-09
DDFFORM 123

LasseStaus:
- Refactor layout variables & move component specfic variables
- Initial mixins for layout-container and vertical-spacing, including variables.
- Added utility classes for layout scss.
- Added documentation and examples for the usage of layout.scss

#### Link to issue

[Jira ticket](https://reload.atlassian.net/browse/DDFFORM-123)

#### Description

This PR implements a starting point for managing the layout of elements/containers and spacing between them for the project. 

Please read the commits, things should mostly be explained there. 

The documentation in the last commit is probably good to read aswell. 

Consider what you think about this as as a whole comment it. This could be anything from explicity of variables to how the mixins are included and used. 

I don't have a clear idea of when/why we would need the utility classes necessarily, but i included it based on our meeting regarding it. 

There is a chance with this layout-container that it could be used excessively inside components. It's primary goal is to container the layout for sections and elements, but not to be the primary way of padding left/right padding to an element and its' children. 

With this in mind, consider if you think this should be specified further or alternative naming, more mixins or something else should be added.

### DIFF
--- a/docs/layout-documentation.md
+++ b/docs/layout-documentation.md
@@ -1,0 +1,138 @@
+# Project Layout documentation & Examples
+
+## Table of Contents
+
+- [Introduction](#introduction)
+- [Variables](#variables)
+- [Mixins](#mixins)
+- [Utility classes](#utility-classes)
+- [Example Usage](#example-usage)
+
+## Introduction
+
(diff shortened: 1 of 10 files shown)


rasben (approved):
Awesome with the documentation.
Looks good 👍

PR #433, opened 2024-01-12
Fix styling

rasben:
Set the page background on the whole body, to avoid margin issues.

Up until now, we've added the page background on various
individual components, but this is a problem if there is
margin between elements, when we want the BG to continue.


LasseStaus (approved):
LGTM. 

However the `<eventParagraphs/>` had two sections. RichText and
 ```   <LinkWithIcon
        href="https://example.com"
        linkText="External Link"
        linkType="external"
      />
      ```
      
The link shouldn't be deleted necesarily i guess. This could be added back.  

I guess the new medias, single and multiple, could be added aswell.

PR #434, opened 2024-01-16
DDFFORM 128 spotlight component

LasseStaus:
- Added recommended-material component.
- Added recommendation component.

#### Link to issue
[jira issue](https://reload.atlassian.net/jira/software/c/projects/DDFFORM/boards/485?selectedIssue=DDFFORM-128)

#### Description

The pr adds the design for the recommendation component & recommended material. 

See commit messages. 

#### Screenshot of the result

material: 
image

Recommendation (wrapper):
image

The arrow is supposed to be large on mobile, as in _the size it is, when hovered_. 

For now i just made the arrow as it is, since the css does not easily support the arrow sgs being modular. This is something i would like to change later on. This isn't a detailed explanation of "why" it doesnt work easily, i'm just mentioning it in case you saw it in the review.

### DIFF
--- a/src/stories/Library/cover/Cover.stories.tsx
+++ b/src/stories/Library/cover/Cover.stories.tsx
@@ -69,7 +69,7 @@ CoverLinked.args = {
   size: "small",
   tint: "80",
   coverUrl: "/",
-  description: "Cover of Audrey Hepburn",
+  alt: "Cover of Audrey Hepburn",
 };
 
 export const SrcNotWork = Template.bind({});
--- a/src/stories/Library/cover/types.ts
+++ b/src/stories/Library/cover/types.ts
@@ -5,6 +5,6 @@ export type CoverProps = {
   tint?: "20" | "40" | "80" | "100" | "120";
   coverUrl?: string;
   alt?: string;
-  shadow?: boolean | "medium";
+  shadow?: "small" | "medium";
   ariaLabel?: string;
 };
--- a/src/stories/Library/cover/types.ts
+++ b/src/stories/Library/cover/types.ts
@@ -4,7 +4,7 @@ export type CoverProps = {
   size: "xsmall" | "small" | "medium" | "large" | "xlarge";
   tint?: "20" | "40" | "80" | "100" | "120";
   coverUrl?: string;
-  description?: string;
+  alt?: string;
   shadow?: boolean;
   ariaLabel?: string;
 };
--- a/src/stories/Library/cover/types.ts
+++ b/src/stories/Library/cover/types.ts
@@ -5,6 +5,6 @@ export type CoverProps = {
   tint?: "20" | "40" | "80" | "100" | "120";
   coverUrl?: string;
   alt?: string;
-  shadow?: boolean;
+  shadow?: boolean | "medium";
   ariaLabel?: string;
 };
--- a/src/stories/Library/recommended-material/RecommendedMaterial.stories.tsx
+++ b/src/stories/Library/recommended-material/RecommendedMaterial.stories.tsx
@@ -7,8 +7,6 @@ export default {
   title: "Library / Recommended Material",
   component: RecommendedMaterial,
   decorators: [withDesign],
-  // We disable the isChecked control, since it is not possible to
-  // get the states from the React component to work with Storybook controls.
   argTypes: {
     favoriteFill: {
       control: "boolean",
--- a/src/stories/Library/recommended-material/RecommendedMaterial.stories.tsx
+++ b/src/stories/Library/recommended-material/RecommendedMaterial.stories.tsx
@@ -16,7 +16,12 @@ export default {
       },
     },
   },
-  parameters: {},
+  parameters: {
+    design: {
+      type: "figma",
+      url: "https://www.figma.com/file/Zx9GrkFA3l4ISvyZD2q0Qi/Designsystem?type=design&node-id=434-6005&mode=design&t=ADgZ8H4KNumis8iG-4",
+    },
+  },
 } as ComponentMeta<typeof RecommendedMaterial>;
 
 const Template: ComponentStory<typeof RecommendedMaterial> = (args) => (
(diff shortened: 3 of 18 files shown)


src/stories/Library/recommended-material/recommended-material.scss
```diff
@@ -0,0 +1,47 @@
+$_max-height-recommended-material: 432px;
+$_aspect-ratio-mobile: 285 / 320;
+$_aspect-ratio-desktop: 378 / 432;
+$_material-image-box-shadow: 0 4px 40px rgba(0, 0, 0, 0.45);
+
+.recommended-material {
+  background-color: $color__global-secondary;
+
+  max-height: 432px;
```
kasperg, line 9:
```suggestion
  max-height: $_max-height-recommended-material;
```

I expect this should use the defined variable?

src/stories/Library/recommended-material/recommended-material.scss
```diff
@@ -0,0 +1,47 @@
...
+  grid-template-columns: 1fr;
+  grid-template-rows: auto 1fr auto;
+  row-gap: $s-xl;
+  aspect-ratio: $_aspect-ratio-mobile;
+  grid-area: material;
+
+  @include media-query__medium {
+    padding: $s-lg;
+    aspect-ratio: $_aspect-ratio-desktop;
+    max-width: 378px;
```
kasperg, line 22: Please consider if this size should be handled differently.

As far as I can tell it is replicated across the recommendation and the recommended material. Can we use a shared variable or at least a comment?

src/stories/Library/recommended-material/RecommendedMaterial.stories.tsx
```diff
@@ -0,0 +1,29 @@
...
+export default {
+  title: "Library / Recommended Material",
+  component: RecommendedMaterial,
+  decorators: [withDesign],
+  // We disable the isChecked control, since it is not possible to
+  // get the states from the React component to work with Storybook controls.
+  argTypes: {
+    isLiked: {
+      control: {
+        disable: true,
```
kasperg, line 15: Appreciate the comment.

kasperg (changes requested):
:+1: I really see you using CSS grid here.

In general please consider if we can reuse existing components. This should make a lot of things easier and more consistent.

Also consider how the components should behave as content expands. You may have to talk to the business about this.

Recommendation - Recommendation Image Left ⋅ Storybook 2024-01-16 17-28-46

src/stories/Library/recommended-material/recommended-material.scss
```diff
@@ -0,0 +1,47 @@
...
+
+  max-height: 432px;
+  width: 100%;
+  padding: $s-md;
+  display: grid;
+  grid-template-columns: 1fr;
+  grid-template-rows: auto 1fr auto;
+  row-gap: $s-xl;
+  aspect-ratio: $_aspect-ratio-mobile;
+  grid-area: material;
```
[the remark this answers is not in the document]
LasseStaus: Good catch, this is a mistake.

src/stories/Library/recommended-material/recommended-material.scss
LasseStaus: If this is in regards to the aspect ratio, i don't think it is used in both the css files. 

The max-width's will be refactored to variables.

src/stories/Library/recommended-material/RecommendedMaterial.tsx
```diff
@@ -0,0 +1,67 @@
...
+              className={
+                isItLiked
+                  ? "bg-identity-primary border-identity-primary"
+                  : "border-secondary-gray"
+              }
+            />
+          </svg>
+        </ButtonUI>
+      </div>
+      <img className="recommended-material__image" src={src} alt={alt} />
```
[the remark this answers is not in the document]
LasseStaus: I've added the cover component instead, and extended it a bit. See in commits.

src/stories/Library/recommended-material/RecommendedMaterial.stories.tsx
```diff
@@ -0,0 +1,33 @@
...
+  component: RecommendedMaterial,
+  decorators: [withDesign],
+  // We disable the isChecked control, since it is not possible to
+  // get the states from the React component to work with Storybook controls.
+  argTypes: {
+    favoriteFill: {
+      control: "boolean",
+      defaultValue: true,
+    },
+  },
```
kasperg, line 17: Please reconsider this comment.

With the latest change it does not seem relevant anymore.

src/stories/Library/cover/Cover.stories.tsx
```diff
@@ -69,7 +69,7 @@ CoverLinked.args = {
   size: "small",
   tint: "80",
   coverUrl: "/",
-  description: "Cover of Audrey Hepburn",
+  alt: "Cover of Audrey Hepburn",
```
kasperg, line 72: I think this is a neat change. It makes it more clear what the prop actually represents.

One might argue that a description could also be used for other purposes within the component but this is not the case here.

src/stories/Library/cover/types.ts
```diff
@@ -4,7 +4,7 @@ export type CoverProps = {
   size: "xsmall" | "small" | "medium" | "large" | "xlarge";
   tint?: "20" | "40" | "80" | "100" | "120";
   coverUrl?: string;
-  description?: string;
-  shadow?: boolean;
+  alt?: string;
+  shadow?: boolean | "medium";
```
kasperg, line 8: Consider changing the type to only be a string. 

With the current setup it can both not be set, be false, be true, or be "medium". An alternate approach would be to make it optional (as now) and then have it accept one of a set of strings e.g. "small", "medium", whatever.

This might also make it easier to generate a control for the value in Storybook.

kasperg (approved):
:+1: Really appreciate the extra work here.

At least update or remove the code comment and then this should be ready to merge.

src/stories/Library/cover/types.ts
LasseStaus: Changed  the old boolean/default to "small" instead.

src/stories/Library/recommended-material/RecommendedMaterial.stories.tsx
LasseStaus: Removed it!

kasperg (approved):
:+1: LGTM

PR #436, opened 2024-01-17
Remove content inner scrollbar from modals

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-142

#### Description
The scrollbar is now only on the very right side of the modal and there isn't any blank space on the bottom of the modal where content is cut off.

#### Screenshot of the result
image

-


spaceo (approved):
Looks good. But are you aware that you also have a change regarding hover and arrows?

PR #439, opened 2024-01-18
Card Grid (nyhedskomponent) element. DDFFORM-127

rasben:
This element is a bit tricky, as the card sizes are very dependent on screen real estate. I've tried to put as much logic as possible in SCSS, so it will be plug-n-play without too much Twig/JSX logic.

### DIFF
--- a/src/stories/Library/card-grid/CardGrid.stories.tsx
+++ b/src/stories/Library/card-grid/CardGrid.stories.tsx
@@ -1,0 +1,71 @@
+import { ComponentStory, ComponentMeta } from "@storybook/react";
+import { withDesign } from "storybook-addon-designs";
+import CardGrid from "./CardGrid";
+import Card from "../card/Card";
+import ImageCredited from "../image-credited/ImageCredited";
+
+const imageUrl =
+  "https://images.unsplash.com/photo-1568667256549-094345857637?q=80&w=2815&auto=format&fit=crop&ixlib=rb-4.0.3&ixid=M3wxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8fA%3D%3D";
+const image = <ImageCredited src={imageUrl} />;
+
+const card = (
+  <Card
+    title="Bøger som har gjort en forskel for romanens udvikling"
+    typeTag="Nyhed"
+    dateTag="06 Dec 2022"
+    image={image}
+  />
+);
+
+const cardNoImage = (
+  <Card
+    title="Fars legestue, hver onsdag"
+    typeTag="Arrangement"
+    dateTag="06 Okt - 28 Dec 2022"
+    placeholderText="Fri leg for alle aldre"
+  />
+);
+
+export default {
+  title: "Library / Card grid ('Nyhedskomponent')",
+  component: CardGrid,
+  decorators: [withDesign],
+  argTypes: {
+    title: {
+      defaultValue: "Nyheder",
+    },
+    linkText: {
+      defaultValue: "Se flere nyheder",
+    },
+    items: {
+      // Disabling controls, as the different card variants are added already.
+      control: false,
+      defaultValue: [card, cardNoImage, card],
+    },
+  },
+  parameters: {
+    design: {
+      type: "figma",
+      url: "Designsystem",
+    },
+  },
+} as ComponentMeta<typeof CardGrid>;
+
+const Template: ComponentStory<typeof CardGrid> = (args) => (
+  <CardGrid {...args} />
+);
+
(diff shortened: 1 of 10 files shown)


src/stories/Library/card-grid/card-grid.scss
```diff
@@ -0,0 +1,129 @@
+// The gap we want between the items.
+// The reason we need to define it explicity is because we need it for a
+// wrapper margin hack.
+$_card-grid-gap: $s-lg;
+
+.card-grid {
```
LasseStaus, line 6: I think this is missing vertical spacing

src/stories/Library/card-grid/CardGrid.stories.tsx
```diff
@@ -0,0 +1,71 @@
+import { ComponentStory, ComponentMeta } from "@storybook/react";
+import { withDesign } from "storybook-addon-designs";
+import CardGrid from "./CardGrid";
+import Card from "../card/Card";
+import ImageCredited from "../image-credited/ImageCredited";
+
+const imageUrl =
```
LasseStaus, line 7: suggestion: 

You could create a file with some different data, instead of using the same. 

I use some different in eventListData.ts, altho i also use the same image. 

Maybe some difference in the design process could be nice since it would be more realistic.

src/stories/Library/card-grid/card-grid.scss
```diff
@@ -0,0 +1,129 @@
...
+  .card-grid--count-5 &,
+  .card-grid--count-6 & {
+    justify-content: end;
+  }
+
+  display: flex;
+  flex-wrap: wrap;
+}
+
+.card-grid__item {
```
LasseStaus, line 53: this is not here in this CSS class, but im jsut marking it here. 

Maybe the approach of using a border for the identity is not the best approach, since they are currently using fixed pixel values. 

For cardList i made "%identity-placeholder-pseudo", which is also using pixels, but i originally thought we could use % of the parent container to place this identity BG on images. 

Personally i think this should be done(or something similar), so small cards with no images, dont have the same huge border like the rest.

LasseStaus (changes requested):
Some different comments you can consider, but i think this needs vertical spacing at least. 

I didn't fully dive into the "width-magic" stuff, but if you think this is the way i'm sure that's correct.  

Consider the image i posted, this is taken from the PR of the cms when i add it to an event, not storybook.

src/stories/Library/card-grid/card-grid.scss
```diff
@@ -0,0 +1,129 @@
...
+    // The design is very specific about the masonry design. See the sizes
+    // defined in the arrays above, which we loop through here.
+    &:nth-child(#{$key}) {
+      width: map_get($_card-widths--small, $name);
+
+      @include media-query__small {
+        width: map_get($_card-widths--medium, $name);
+      }
+
+      @include media-query__large {
```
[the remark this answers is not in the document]
rasben: There's no images, cause this article has no `teaser image` nor `teaser text` added.

Also, it appears that you've tested it against an old version of release/2024-04 - the grid should not have a max-width.

src/stories/Library/card-grid/CardGrid.stories.tsx
rasben: I get what you're saying, but, I think this is something we should look into in a more systematic and broader way.

src/stories/Library/card-grid/card-grid.scss
rasben: I completely agree with you, but the design marks the borders based on PX values.
I agree with you that percentage would look better, but I think it's something we should take as part of the 'the teasers are very big' conversation on monday :)

src/stories/Library/card-grid/card-grid.scss
rasben: Nope, the vertical spacing comes from `.paragraph` - if card-grid stands completely alone (outside CMS), it shouldnt have bottom margin.

https://github.com/danskernesdigitalebibliotek/dpl-design-system/blob/release/2024-4-0/src/stories/Library/paragraph/paragraph.scss

src/stories/Library/card-grid/card-grid.scss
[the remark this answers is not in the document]
LasseStaus: For this branch i still have 
```
.paragraph {
  @include layout-container;
  @include vertical-spacing;
}

```
in the css? 

I guess this is the thing that should be / already is changed in another pr?

LasseStaus (approved):
LGTM if you say the paragraph max width is fixed

PR #443, opened 2024-01-21
Optimize github action ressource usage.

rasben:
- Only run the most ressource heavy jobs on non-draft PRs. Stuff like Chromatic deployment only makes sense once the pull request has actually been marked as ready for review, as other times, you might push several times to the branch
- Add a concurrency checker for all jobs, cancelling on-going jobs if a new push to the same branch has been detected.


rasben:
Related: https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/668

spaceo (approved):
LGTM 👍

.github/workflows/create-release-on-branch-changes.yml
```diff
@@ -6,6 +6,12 @@ on:
       # Dependabot does not have permissions to create releases
       - 'dependabot/**'
 
+# Detect if this action is already running, and cancel it.
+# This most likely happened because a second push has been made to a branch.
+concurrency:
+  group: ${{ github.repository_id }}-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
```
kasperg, line 14: I wonder if this is wise.

I fear we end up in situations where a job is half done e.g. created as GitHub release but not as an NPM package.

.github/workflows/chromatic.yml
```diff
@@ -1,9 +1,28 @@
 name: "Chromatic"
 
-on: push
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    types: [ready_for_review, opened, synchronize]
```
kasperg, line 8: I wonder if this is wise.

1. We should at least add `develop` as a branch we check for when pushing. It is the current main branch. Also: What about `release/*` branches?
2. What happens if someone creates a PR against a branch not covered by any of the above? Will Chromatic have a proper baseline to test against?

.github/workflows/chromatic.yml
```diff
@@ -1,9 +1,28 @@
...
+  group: ${{ github.repository_id }}-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   chromatic-deployment:
+
+    # Don't run action if the pull request is a draft.
+    # This is an expensive job to run, so we only want to run it when the
+    # PR is actually ready to be reviewed.
+    if: ${{ github.event.pull_request.draft == false }}
```
kasperg, line 22: My comment for the CMS also applies here: https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/668#discussion_r1467802721

PR #446, opened 2024-01-22
Button for paying fees in header on  /user/me/fees page

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-368

#### Description
These changes:
* Adds top margin to the fees list to add more air to the top interface
* Adds a fee list story (**with introduced external link button**) so we can actually work with the view

#### Screenshot of the result

image


Dresse (approved):
LGTM

PR #447, opened 2024-01-22
Card+Card-grid size tweaks. DDFFORM-180 DDFFORM-182

rasben:
Adjust card-grid max width, to avoid massive cards. DDFFORM-180

---

Fix cards sizing, and slider 'jumping' DDFFORM-182

- Make every second item in slider have margin-top, to pull it down, instead of based on no-image.
- Add missing mobile styling for tall (medium) cards.
- Add framing to XL card teasers with images
- Adjust the framing and padding on cards, to make them more sensible on smaller screens.


kasperbirch1 (approved):
LGTM 👍

PR #451, opened 2024-01-24
Use same arrow button behavior for list materials cards as other cards

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-388

#### Description
All arrows need to behave the same across the apps

#### Screenshot of the result

https://github.com/danskernesdigitalebibliotek/dpl-design-system/assets/28546954/690a07a9-aa3a-41e7-81cc-1034ade7b979

-

### DIFF
--- a/src/stories/Library/Lists/list-materials/list-materials.scss
+++ b/src/stories/Library/Lists/list-materials/list-materials.scss
@@ -1,5 +1,7 @@
 .list-materials {
   @include show-svg-on-hover(".list-dashboard__arrow");
+  @include show-svg-on-hover(".list-materials__arrow");
+  @include show-svg-on-hover(".arrow-button");
 
   display: flex;
   min-height: 96px;


src/stories/Library/Lists/list-materials/list-materials.scss
```diff
@@ -1,5 +1,7 @@
 .list-materials {
   @include show-svg-on-hover(".list-dashboard__arrow");
+  @include show-svg-on-hover(".list-materials__arrow");
+  @include show-svg-on-hover(".arrow-button");
```
spaceo, line 5: Without having dived into the definitions of `show-svg-on-hover(".list-materials__arrow")` and `show-svg-on-hover(".arrow-button")`: Does it make sense to have both. Couldn't it be handled with only one definition?

Adamik10: Short answer: no.
Long answer: The DOM structure can vary between the different types of cards we use, and the container in which the arrow component is located can have any of these classes. So we need to make sure we target them with `show-svg-on-hover()`

spaceo (approved):
LGTM 👍

PR #452, opened 2024-01-24
Taxonomy categories and tags

kasperg:
#### Link to issue

[Please add a link to the issue being addressed by this change.
](https://reload.atlassian.net/browse/DDFFORM-130)

#### Description

This PR adds support for category and tags for events and articles.

A notable change is the rendering of categories. According to Figma they should use the large tag component on the full view of events and articles.

#### Screenshot of the result

Check out Chromatic.

### DIFF
--- a/src/stories/Blocks/article/Article.stories.tsx
+++ b/src/stories/Blocks/article/Article.stories.tsx
@@ -23,6 +23,16 @@ export default {
     date: {
       defaultValue: "08. April 21",
     },
+    tags: {
+      defaultValue: [
+        "dans",
+        "contemporary",
+        "modern",
+        "scenekunst",
+        "digt",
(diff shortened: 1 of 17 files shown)


LasseStaus (approved):
Looks good and definitely better with these separated tag variations. 

Personal take: I would remove the line-height on the tags. The non-button versions are slightly off vertical center because of it. 
Currently it is 120%. If this is intended then just keep it as is. 

Other than that it looks good to me!

kasperg:
> Personal take: I would remove the line-height on the tags. The non-button versions are slightly off vertical center because of it. Currently it is 120%. If this is intended then just keep it as is.

:+1: I see it as intended as they use the same component as elsewhere.

PR #463, opened 2024-01-30
Nav-grid (Navigationskomponent) - DDFFORM-164

rasben:
A new component, with the same concept as `card-grid` (header, items etc.)

Also includes some simple Javascript, that 'folds out' the list of items if there are more than 6.

https://reload.atlassian.net/browse/DDFFORM-164

You can see it in action here:

https://616ffdab9acbf5003ad5fd2b-adqwgitnrd.chromatic.com/?path=/story/library-nav-grid-navigationskomponent--many


src/stories/Library/nav-grid/nav-grid.scss
```diff
@@ -0,0 +1,74 @@
+.nav-grid {
+  @include layout-container($layout__max-width--large, 0);
+  @include vertical-spacing();
+}
+
+.nav-grid--simple {
+  .nav-teaser__subtitle {
+    display: none;
+  }
+}
```
LasseStaus, line 10:
```suggestion
.nav-grid--simple.nav-teaser__subtitle {
   display: none;
  }
```

i guess no reason to nest

src/stories/Library/nav-teaser/NavTeaser.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+        <svg
+          xmlns="http://www.w3.org/2000/svg"
+          width="33"
+          height="8"
+          viewBox="0 0 33 8"
+          fill="black"
+        >
+          <path d="M32.287 4.35356C32.4823 4.15829 32.4823 3.84171 32.287 3.64645L29.105 0.464469C28.9098 0.269206 28.5932 0.269206 28.3979 0.464469C28.2027 0.659731 28.2027 0.976313 28.3979 1.17158L31.2264 4L28.3979 6.82843C28.2027 7.02369 28.2027 7.34027 28.3979 7.53554C28.5932 7.7308 28.9098 7.7308 29.105 7.53554L32.287 4.35356ZM-4.37114e-08 4.5L31.9335 4.5L31.9335 3.5L4.37114e-08 3.5L-4.37114e-08 4.5Z" />
+        </svg>
```
LasseStaus, line 24: I think we should use `  <ArrowSmallRight />` or similar to avoid these large svg pastes.

LasseStaus (approved):
LGTM, nice one. 

I commented a question, and made two suggestions i think you should change. 

🥳

src/stories/Library/nav-grid/nav-grid.scss
rasben: the reason to nest is to make it easier to read :)

src/stories/Library/nav-teaser/NavTeaser.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+const NavTeaser: FC<NavTeaserProps> = ({ title, subtitle }) => {
+  return (
+    <article className="nav-teaser">
+      <a href="#">
+        <h1 className="nav-teaser__title">{title}</h1>
+
+        {subtitle ? <p className="nav-teaser__subtitle">{subtitle}</p> : ""}
+
+        <svg
```
LasseStaus, line 16: The arrow is not growing on hover. I guess that this should happen same as everywhere else

src/stories/Library/nav-teaser/NavTeaser.tsx
```diff
@@ -0,0 +1,30 @@
+import { FC } from "react";
+
+type NavTeaserProps = {
+  title: string;
+  subtitle?: string;
+};
+
+const NavTeaser: FC<NavTeaserProps> = ({ title, subtitle }) => {
+  return (
+    <article className="nav-teaser">
```
LasseStaus, line 10: Was it decided to remove the pagefold on hover? 

image

LasseStaus (changes requested):
See comments, found some things after viewing the CMS

I think you also could add this component to the block story for Article or similar.

rasben:
>I think you also could add this component to the block story for Article or similar.

This will be done by KasperB

src/stories/Library/nav-grid/NavGrid.tsx
```diff
@@ -0,0 +1,47 @@
+import { FC, ReactNode, useEffect } from "react";
+import clsx from "clsx";
+
+type CardProps = {
+  title?: string;
+  showSubtitles?: boolean;
```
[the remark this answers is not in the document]
rasben: I dont understand this comment - right now, it is an optional value (e.g. `?`)

src/stories/Library/nav-teaser/NavTeaser.tsx
rasben: Good point - I've added it now. It didnt exist already.

src/stories/Library/nav-teaser/NavTeaser.tsx
rasben: It's also a totally different size than all the other arrows.

src/stories/Library/nav-teaser/NavTeaser.tsx
LasseStaus: I see yea

src/stories/Library/nav-teaser/NavTeaser.tsx
LasseStaus: If you say that we agreed on the meeting that this specific component should not have hover behaviour, then that is fine. However, i don't remember this myself and i find it strange that this should behave differently. 

If you prefer not to, then sure, if it was me i would add it.

src/stories/Library/nav-grid/NavGrid.tsx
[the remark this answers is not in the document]
LasseStaus: I'm asking wether it should be possible to decide on the individual item, whether it should include subtitle or not, rather than control this on the parent. 

We'll just take it at demo. 

I guess this can be accomplished regardless right now, by allowing subtitles and just not filling a description for the individual item. 

So it does work both ways already in that sense.

LasseStaus (approved):
LGTM 👍

PR #467, opened 2024-01-31
Add .npmrc file in release workflows

spaceo:
Apparently it works better to authenticate the npm read/write with a .npmrc file.


Adamik10 (approved):
APPROVED!!!

PR #470, opened 2024-02-02
Hover drop shadow on fee list items

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-394

#### Description

Add drop shadow on fee items on hover.

I also refactored the card list page story and involved components to "Search Result Page".
Because that is what it is. When I started the task I had a hard time locating that story because of that.


clausbruun (approved):
Looks good and it's awesome with the refactoring.

PR #471, opened 2024-02-02
Allow logos to also support images without height/width set. (E.g. SVG)

rasben:
DDFFORM-176 DDFFORM-177

Relies on:

https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/717


kasperbirch1 (approved):
LGTM 👍

PR #473, opened 2024-02-02
Update faulty line-height, to not have broken double-lines.

rasben:
42px appears to be some kind of typo - either way, a line-height of 42px when using 16px font-size is extreme and makes no sense. Let's use the 160% that is already being used for `small` instead.


LasseStaus (approved):
Agreed with the changes to the line height. 
Whether we used 160% or another fixed value i'll leave up to you, but the current LH is definitely not good.

LGTM. Remember to accept.

PR #476, opened 2024-02-02
update small caption font size

LasseStaus:
- Changing the font-size for `small text caption` because it is too small too meet accessbility for the places it is used.

#### Link to issue

[Jira link](https://reload.atlassian.net/jira/software/c/projects/DDFFORM/boards/485?assignee=557058%3A9a3dd056-14db-4baf-917f-d89c6c42dceb&selectedIssue=DDFFORM-174)

#### Description

This PR changes the base font-size of 'Small text caption" from 10px to 12px. 

The font is used across the system, but it is too small for optimal readability. 

I've left the chromatic changes unaccepted on purpose, so that the review can look through them and see what they think.


kasperbirch1 (approved):
It looks fine to me, but as you yourself suggested, it might be a good idea for the React team to also look it over.

spaceo (commented):
Actually I think we should have DDF veridying these changes.
Some og the mobile views gets quite crammed:

https://www.chromatic.com/test?appId=616ffdab9acbf5003ad5fd2b&id=65c361e04e11c69f9270ff31

https://www.chromatic.com/test?appId=616ffdab9acbf5003ad5fd2b&id=65c361e04e11c69f9270ff53

Most emails will break down on two lines here:
https://www.chromatic.com/test?appId=616ffdab9acbf5003ad5fd2b&id=65c361e04e11c69f9270ff5d

LasseStaus:
Superseded by #536

PR #482, opened 2024-02-05
Add loan page story and skeletons

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-429
https://reload.atlassian.net/browse/DDFLSBP-430

#### Description
This PR introduces the Loan page story with multiple substories:
- default
- no physical loans
- no digital loans
- no loans at all
- stacked loans
- skeleton

You might have noticed that this PR has a bunch of changes - that's because as a part of this PR we change the way we use SVGs in the repository. Now we use them in the same way as in the dpl-react repository - importing them as React components, as opposed to using a given svg as an image src attribute. 

#### Screenshot of the result
image

-


src/stories/Library/warning-status/warning-status.scss
```diff
@@ -11,6 +11,7 @@
 
 .warning-bar__icon {
   margin-right: $s-md;
+  min-width: 40px;
```
spaceo, line 14: I think we should use spacing variable here.

spaceo (changes requested):
Great work Adam!
A lot of effort went into this!
I have one remark to this piece of art :)

src/stories/Library/warning-status/warning-status.scss
spaceo: hmmm move along nothing to see here...

spaceo (approved):
LGTM 👍

PR #485, opened 2024-02-06
Fix tag categories

kasperbirch1:
#### Depend on:
https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/733

#### Description

This PR accomplishes two things: 
- Rename `TagCategories` to `TagList`. 
- Replaces `TagList` with `Tag` in instances where `TagList` is not necessary.


LasseStaus (approved):
LGTM! Nice one.

PR #487, opened 2024-02-06
Setup 'categories' exposed filter on event search page. DDFFORM-184

rasben:
A few minor updates, displaying the dropdown on the event-list-page element.

Also introduces a few small fixes, such as padding on the dropdown.

Easiest to review commit by commit.

See it here:

https://616ffdab9acbf5003ad5fd2b-fakpuatovd.chromatic.com/?path=/story/blocks-event-list-page--default

### DIFF
--- a/src/stories/Blocks/event-list-page/EventListPage.tsx
+++ b/src/stories/Blocks/event-list-page/EventListPage.tsx
@@ -1,11 +1,30 @@
 import React from "react";
 import EventList from "../../Library/event-list/EventList";
 import eventListData from "../../Library/event-list/EventListData";
+import { InputLabel } from "../../Library/input-label/InputLabel";
+import { Dropdown } from "../../Library/dropdown/Dropdown";
+
+const list = [
+  {
+    title: "Bob Dylan 80 år",
+    href: "/",
+  },
+  {
+    title: "5 Nordiske fuldtræffere",
+    href: "/",
+  },
+];
 
 const EventListPage: React.FC = () => {
   return (
     <div className="event-list-page">
       <h1 className="event-list-page__heading">Arrangementer</h1>
+      <div className="event-list-page__filters">
+        <div className="event-list-page__filter">
+          <InputLabel text="Kategorier" />
+          <Dropdown list={list} ariaLabel="Kategorier" arrowIcon="chevron" />
+        </div>
+      </div>
       <EventList events={eventListData} />
     </div>
   );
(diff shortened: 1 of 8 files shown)


src/stories/Library/dropdown/dropdown.scss
```diff
@@ -17,18 +19,22 @@
...
   cursor: pointer;
   background-color: transparent;
   color: $color__global-black;
   border-radius: 0;
   border: 1px solid;
   border-color: inherit;
-  padding: 0 50px 0 20px;
+  padding: 0;
+  padding-left: $_padding;
+  padding-right: $_dropdown-icon-size + $_padding;
```
kasperg, line 37: As I see it this will change `padding-right` from 50px to 70px (50 + 20). I wonder if that is intentional?

src/stories/Library/dropdown/dropdown.scss
```diff
@@ -1,3 +1,5 @@
+$_dropdown-icon-size: 50px;
```
kasperg, line 1: :+1: Nice refactoring.

src/stories/Blocks/event-list-page/EventListPage.tsx
```diff
@@ -1,11 +1,30 @@
...
+];
 
 const EventListPage: React.FC = () => {
   return (
     <div className="event-list-page">
       <h1 className="event-list-page__heading">Arrangementer</h1>
+      <div className="event-list-page__filters">
+        <div className="event-list-page__filter">
+          <InputLabel text="Kategorier" />
+          <Dropdown list={list} ariaLabel="Kategorier" arrowIcon="chevron" />
```
kasperg, line 25: Please reconsider this construct.

As I see it the label value is duplicated between the label component and the Dropdown component.

If we have an explicit label I would prefer that we had an explicit link between the two elements using `for` on the `<label>` or [`aria-labelledby`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby) on the form element.

kasperg (approved):
:+1: I will accept this for now to avoid blocking but I think there is some elements here that require serious consideration.

Note that this means changes to the collect modal as seen in Chromatic. I do not think that is intentional?

src/stories/Library/dropdown/dropdown.scss
rasben: It is intentional :) It appears to be broken as it is.

PR #489, opened 2024-02-07
Group modal card skeletons

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-427

#### Description
This PR introduces group modal item skeletons with styling and a brand new story.
As a last minute decision, I added dashboard list item skeletons too - to be used on the dashboard and in the user menu.

#### Screenshot of the result
image

image

-

### DIFF
--- a/src/stories/Library/Lists/list-dashboard/list-dashboard-skeleton.scss
+++ b/src/stories/Library/Lists/list-dashboard/list-dashboard-skeleton.scss
@@ -1,0 +1,20 @@
+// We don't have any of these dimensions in variables, but we want to preserve
+// the skeleton components' dimensions the same as the actual components.
+.list-dashboard.ssc {
+  height: 72px;
+
+  @include media-query__small {
+    height: 88px;
+  }
+
+  & .ssc-circle {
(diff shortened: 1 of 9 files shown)


src/stories/Library/Lists/list-dashboard/list-dashboard-skeleton.scss
```diff
@@ -0,0 +1,20 @@
+// We don't have any of these dimensions in variables, but we want to preserve
+// the skeleton components' dimensions the same as the actual components.
```
spaceo, line 2: 👍 For comment!

spaceo (changes requested):
Why not having skeleton screens in the header of the modal?
As far as I can see we have external data there as well.
image

Adamik10:
@spaceo I have been testing it with the lowest possible network throttling option selected (in react), and those parts of the modal must already be loaded when the modal is created. So I didn't want to spend time on something that wouldn't really make a difference in the end.

spaceo (approved):
LGTM 👍

PR #491, opened 2024-02-07
Make WYSIWYG.css file, for use in CKEditor injection. DDFFORM-241

rasben:


kasperg (approved):
:+1: Very nice.

Two comments for you consideration but nothing blocking.

PR #493, opened 2024-02-08
Refactor header, to automatically enable burger-menu when needed.

rasben:
`.has-burger-menu` is added dynamically using JS, by calculating if there is enough space to show the desktop menu, and otherwise enabling the burger menu.
This required some re-factoring of the header menu system, as I need to measure the width of the desktop menu at all times. So, instead of display: none, we're just hiding it visually.

DDFFORM-220

### DIFF
--- a/src/stories/Blocks/autosuggest/Autosuggest.stories.tsx
+++ b/src/stories/Blocks/autosuggest/Autosuggest.stories.tsx
@@ -25,42 +25,12 @@ export default {
 const Template: ComponentStory<typeof Autosuggest> = (
   args: AutosuggestProps
 ) => (
-  // Apart from <Autosuggest />, everything else is here just for the story
-  // context. The autosuggest styling is directly dependent on the header component,
-  // Please keep up to date with Header.tsx
-  <header className="header" style={{ height: "144px" }}>
-    <div className="header__logo-desktop">
-      <p className="text-body-medium-regular">Context</p>
-    </div>
-    <div className="header__menu">
-      <nav className="header__menu-first">
-        <p className="text-body-medium-regular">Context</p>
-      </nav>
-      <div className="header__menu-second">
-        <div className="header__menu-search">
-          <input
-            className="header__menu-search-input text-body-medium-regular"
-            type="text"
-            placeholder="This field is here just for context."
-          />
-          <img
-            className="header__menu-search-icon"
-            src="icons/basic/icon-search.svg"
-            alt="search icon"
-          />
-          <img
-            className="header__menu-dropdown-icon"
-            src="icons/collection/ExpandMore.svg"
-            alt="expand dropdown icon"
-          />
-          <Autosuggest {...args} />
-        </div>
-      </div>
-    </div>
-    <div className="header__clock">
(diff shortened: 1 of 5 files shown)


src/stories/Blocks/autosuggest/Autosuggest.stories.tsx
```diff
@@ -25,42 +25,9 @@ export default {
...
-          />
-          <Autosuggest {...args} />
-        </div>
-      </div>
-    </div>
-    <div className="header__clock">
-      <p className="text-body-medium-regular">Context</p>
-    </div>
-  </header>
+  <div style={{ position: "relative" }}>
```
kasperbirch1, line 28: Is there a specific reason to use inline style? I think this is something we would like to avoid.

src/stories/Blocks/autosuggest/Autosuggest.stories.tsx
```diff
@@ -25,42 +25,9 @@ export default {
 const Template: ComponentStory<typeof Autosuggest> = (
   args: AutosuggestProps
 ) => (
-  // Apart from <Autosuggest />, everything else is here just for the story
-  // context. The autosuggest styling is directly dependent on the header component,
-  // Please keep up to date with Header.tsx
-  <header className="header" style={{ height: "144px" }}>
```
kasperbirch1, line 31: Is it necessary to remove this? As I see it, autosugest needs to be in the header to give the right impression

src/stories/Blocks/header/header-state.js
```diff
@@ -0,0 +1,47 @@
...
+    document.documentElement.classList.remove("has-desktop-menu");
+  } else {
+    document.documentElement.classList.remove("has-burger-menu");
+    document.documentElement.classList.add("has-desktop-menu");
+  }
+}
+
+// Function for initializing the header state based on window size and container width.
+function initHeaderState() {
+  // Finds the node with class 'header__menu-navigation' which is not initialized.
```
kasperbirch1, line 25: Cool featue! But shouldn't you wrap it in a document.addEventListener("DOMContentLoaded", () => {} to make sure the markup is ready?

kasperbirch1 (approved):
I have asked some questions but otherwise cool feature 👍

src/stories/Blocks/header/header-state.js
rasben: Good point, done 👍

src/stories/Blocks/autosuggest/Autosuggest.stories.tsx
rasben: I get what you're saying, but yes, it is necessary.
I've added a comment to make it more clear

src/stories/Blocks/autosuggest/Autosuggest.stories.tsx
rasben: As I write in the commit:

> 
> Update autosuggest story to not be dependent on header.
> It's not maintainable nor necessary for us to display the whole
> header in the Autosuggest story.
> If it is necessary in the future, we'll have to make the header
> element a proper element, that can take the auto-suggest element.

PR #494, opened 2024-02-09
Fix iOS 15 default blue color and add background color to close icon [DDFLSBP-414]

clausbruun:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-414

#### Description

* Adding a background color to the close icon in modals to prevent interference with the content in the modal.
* Setting a specific color for texts that would otherwise adopt the default iOS 15 blue color.
* Transferring changes made to the user icon in the header from dpl-react to the dpl-design-system.

#### Screenshot of the result

_Close icon background color:_

https://github.com/danskernesdigitalebibliotek/dpl-design-system/assets/105956/9d584289-2c3f-4073-be34-c70e0d0352e0

_No blue texts:_

image

image

*


src/stories/Library/Lists/list-reservation/list-reservation.scss
```diff
@@ -18,6 +18,14 @@ $list-reservation-space-desktop: 24px;
   padding: $list-reservation-space-mobile;
   cursor: pointer;
 
+  h3 {
+    color: $color__text-primary-black;
```
Adamik10, line 22: Very nice that you're using the variables 👏 💯

src/stories/Blocks/header/Header.tsx
```diff
@@ -109,21 +109,19 @@ export const Header = (props: HeaderProps) => {
...
-                {signedIn && haveNotification && (
-                  <div className="header__notification bg-signal-alert" />
-                )}
-                <ProfileIcon />
-                {signedIn && (
-                  <span className="text-small-caption">{username}</span>
-                )}
-              </a>
-            </div>
+            <button
```
Adamik10, line 112: 👏 Great that you're bringing the actual implementation back to the design system 👏

Adamik10 (approved):
Looks perf! You just need to check out your Chromatic UI tests. We need to talk to Kasper G if you don't have the rights to approve them (note that you have to sign in with. your GH user in there)

src/stories/Library/Lists/list-reservation/list-reservation.scss
clausbruun: Thank you @Adamik10 🙏🏻

clausbruun:
> The bottom border doesn't align with the rest of the borders now. Do you know why that could be? I would recommend that to be solved before we move this forward 🙈

I'm sorry, I should have caught that myself. 🫠

clausbruun:
The border property was [inherited](https://github.com/danskernesdigitalebibliotek/dpl-design-system/blob/64108bc1ebfea865e123604274e100a03902ce91/src/styles/scss/reset.scss#L72) from [this element](https://github.com/danskernesdigitalebibliotek/dpl-design-system/blob/64108bc1ebfea865e123604274e100a03902ce91/src/stories/Blocks/header/header.scss#L45), so I have added a fix for that.

PR #502, opened 2024-02-15
Custom movemenet calculator for sliders, when using keyboard tabbing.

rasben:
SwiperJS has a hard time understanding how to deal with keyboard focus and setting the slider positioning.
Because we have items with very different widths, and we use "auto" for the actual movement, the focus goes out of sync with the visual visibility. This function is called when the swiper is initialized. We listen on focus within the item wrapper, and when we detect it, we calculate our own transform value, overwriting SwiperJS.

DDFFORM-233

### DIFF
--- a/src/stories/Library/slider/init-slider.js
+++ b/src/stories/Library/slider/init-slider.js
@@ -1,11 +1,60 @@
+// SwiperJS has a hard time understanding how to deal with keyboard focus and
+// setting the slider positioning.
+// Because we have items with very different widths, and we use "auto" for
+// the actual movement, the focus goes out of sync with the visual visibility.
+// This function is called when the swiper is initialized.
+// We listen on focus within the item wrapper, and when we detect it, we
+// calculate our own transform value, overwriting SwiperJS.
+function swiperWrapperEventInit(swiperWrapper) {
+  swiperWrapper.addEventListener("focusin", () => {
+    const activeSlide = swiperWrapper.querySelector(".swiper-slide-active");
+
+    if (!activeSlide) {
+      return;
+    }
+
+    let currentSibling = activeSlide;
+    let translateWidth = 0;
+    let sideMargins = 0;
+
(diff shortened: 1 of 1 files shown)


src/stories/Library/slider/init-slider.js
```diff
@@ -1,11 +1,60 @@
...
+    }
+
+    // Calculate a new translate value, for pulling the slider.
+    // We add half of the latest known sidemargin, so the active slide doesn't
+    // go flush against the screen.
+    // eslint-disable-next-line no-param-reassign
+    swiperWrapper.style.transform = `translate3d(${Math.floor(
+      translateWidth + sideMargins / 2
+    )}px, 0, 0)`;
+  });
```
kasperg, line 41: So this is a bit crazy.

I expect that you have tried [`swiper.slideTo()`](https://swiperjs.com/swiper-api#method-swiper-slideTo) and this did not work.

Have you considered calling [swiper.translateTo()](https://swiperjs.com/swiper-api#method-swiper-translateTo) instead of setting the style directly?

kasperg (approved):
:+1: This is somewhat crazy but if this is what it takes...

src/stories/Library/slider/init-slider.js
rasben: >I expect that you have tried [swiper.slideTo()](https://swiperjs.com/swiper-api#method-swiper-slideTo) and this did not work.

Oh yes, dont get me started on that rabbit hole..

>Have you considered calling [swiper.translateTo()](https://swiperjs.com/swiper-api#method-swiper-translateTo) instead of setting the style directly?

Great idea. I'll try to see if it'll work, and otherwise, we'll stick with this.

rasben: I've tried it, but, sadly translateTo() messes up the actual transitioning. Sigh.

PR #504, opened 2024-02-15
Dont set focus to first element in slider when tabbing, unless visible.

rasben:
When using the keyboard, and using the arrow controls, there is a mismatch when you tab from the controls into the actual slider. It always wants to add the focus to the first item in the slider, meaning the slider will be reset.
We want to instead set focus on the first fully-visible slide. However, we obviously don't want to set the focus on this, if you're moving backwards (e.g. from slider => controls => whatever). We get around this, by creating a "focus controller" - e.g. a custom element that is created after the controls.
When focus is introduced to the focus controller, it gets removed, and we set the focus to the first active slide.
We create the FocusController whenever the focus is set on the controls, so that way, it's only when moving 'forwards' in the DOM tree that we hijack focus.

DDFFORM-232

### DIFF
--- a/src/stories/Library/slider/init-slider.js
+++ b/src/stories/Library/slider/init-slider.js
@@ -1,3 +1,63 @@
+// When using the keyboard, and using the arrow controls, there is a mismatch
+// when you tab from the controls into the actual slider.
+// It always wants to add the focus to the first item in the slider,
+// meaning the slider will be reset.
+// We want to instead set focus on the first fully-visible slide.
+// However, we obviously don't want to set the focus on this, if you're moving
+// backwards (e.g. from slider => controls => whatever).
+// We get around this, by creating a "focus controller" - e.g. a custom element
+// that is created after the controls.
+// When focus is introduced to the focus controller, it gets removed, and we set
+// the focus to the first active slide.
+// We create the FocusController whenever the focus is set on the controls, so
+// that way, it's only when moving 'forwards' in the DOM tree that we hijack
+// focus.
+function createFocusController(swiper, controls) {
+  const focusControllerClass = "slider__focus-controller";
+  let focusController = controls.querySelector(`.${focusControllerClass}`);
+
+  // If a focus controller already exists, we won't create a new one.
+  if (focusController) {
+    return;
+  }
+
+  // Creating our custom focus element.
+  focusController = document.createElement("button");
+  focusController.classList.add(focusControllerClass);
+  controls.appendChild(focusController);
+
+  // When focus is set to the controller, we find the first visible slide,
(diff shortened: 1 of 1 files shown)


src/stories/Library/slider/init-slider.js
```diff
@@ -1,3 +1,63 @@
...
+  focusController = document.createElement("button");
+  focusController.classList.add(focusControllerClass);
+  controls.appendChild(focusController);
+
+  // When focus is set to the controller, we find the first visible slide,
+  // and set the focus to that.
+  // We also remove the focus controller, to make sure that we can tab
+  // backwards up the tree, without being hijacked.
+  focusController.addEventListener("focus", () => {
+    focusController.remove();
```
kasperg, line 34: From an accessibility perspective I worry about this approach. It creates an (invisible?) button without without any content/text. I wonder how this will behave for screen readers and the like.

kasperg (approved):
I find it hard to determine if this is the right approach to solving this. To me it seems like a giant hack.

It might be turning the problem on its head but I wonder if it would be more appropriate to remove the possibility of using the UI arrows with the keyboard. If a visitor wants to move through the slider they can use tab/reverse-tab which provides the exact same functionality. With #502 we also ensure that visibility of items in the slider follow focus.

src/stories/Library/slider/init-slider.js
rasben: I get what you're saying, but, technically speaking, the screenreader wont even detect it, as, as soon as you focus it, it gets removed.

I'll ask Nino & co to be extra dilligent when testing the sliders, both in accessibility and general experience.

src/stories/Library/slider/init-slider.js
```diff
@@ -7,6 +67,9 @@
 // calculate our own transform value, overwriting SwiperJS.
 function swiperWrapperEventInit(swiperWrapper) {
   swiperWrapper.addEventListener("focusin", () => {
+    // eslint-disable-next-line no-param-reassign
+    swiperWrapper.style.transitionDuration = "300ms";
```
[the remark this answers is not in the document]
rasben: I'll add a comment - but, basically, SwiperJS really wants to remove the tranisitionduraton when not in use, and if i "hardcode" it in CSS, the swiper acts up majorly when using a finger to move it.

rasben:
>It might be turning the problem on its head but I wonder if it would be more appropriate to remove the possibility of using the UI arrows with the keyboard.

Derp. Good idea, actually. I think I got so lost in the idea that being visually available, the arrows should also be focusable.

rasben:
ja

PR #505, opened 2024-02-15
Don't capitalize all words in the counter label text, only first letter

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-409

#### Description
We now only capitalize the first letter of the counter label text, as opposed to every word.

#### Screenshot of the result
-

-


Dresse (approved):
LGTM

PR #506, opened 2024-02-15
Add stack display of event series

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-297

#### Description
This pull request introduces a new component, `EventListItemStacked`. The primary function of this component is to manage recurring events, providing a user-friendly interface that clearly displays the events occurring on different dates.

#### Screenshot of the result
image
**Tablet**
image
**Mobile**
image


src/stories/Library/event-list-item/EventListItemStacked.stories.tsx
```diff
@@ -0,0 +1,36 @@
...
+    },
+    href: {
+      defaultValue: "/",
+      control: { type: "text" },
+    },
+  },
+  parameters: {
+    design: {
+      type: "figma",
+      url: "",
```
LasseStaus, line 26: Missing link if this is part of figma.

src/stories/Library/event-list-item/event-list-item.scss
```diff
@@ -1,21 +1,64 @@
...
+  &::before {
+    content: "";
+    background-color: $color__global-white;
+    position: absolute;
+    width: calc(100% - $_stacked-reduce-width);
+    height: 100%;
+    left: 50%;
+    top: 0;
+    transform: translateX(-50%);
+    z-index: -1;
```
LasseStaus, line 18:
```suggestion
```

I'd avoid using - values for this, and instead add a higher z-index for -stacked__content

src/stories/Library/event-list-item/event-list-item.scss
```diff
@@ -100,27 +141,52 @@ $_desktop-content-rows: 5;
...
+  padding-top: $s-sm;
+  padding-bottom: $s-sm;
+  grid-template-areas: "contentStacked";
+  align-items: center;
+}
+
+.event-list-item-stacked__content {
+  @include typography($typo__body-placeholder);
+  @include event-content-layout;
+  grid-area: contentStacked;
```
LasseStaus, line 159:
```suggestion
z-index: 10; (with variable)
  grid-area: contentStacked;
```

src/stories/Library/event-list-item/event-list-item.scss
```diff
@@ -100,27 +141,52 @@ $_desktop-content-rows: 5;
...
+  grid-area: dateStacked;
+}
+
+.event-list-item-stacked__time {
+  grid-area: timeStacked;
+  justify-self: end;
+}
+
+.event-list-item-stacked > svg {
+  grid-area: arrowStacked;
```
LasseStaus, line 173:
```suggestion
z-index: 10; (with variable)
  grid-area: arrowStacked;
```

src/stories/Library/event-list-item/event-list-item.scss
```diff
@@ -1,21 +1,64 @@
...
+    height: 100%;
+    left: 50%;
+    top: 0;
+    transform: translateX(-50%);
+    z-index: -1;
+    box-shadow: inset 0 15px 15px -10px rgba(0, 0, 0, 0.1);
+  }
+}
+
+@mixin event-layout {
```
LasseStaus, line 23: Nice with the mixin, i'll just add some name suggestions you can consider if you want or not. 


```suggestion
@mixin event-list-item-grid {
```

src/stories/Library/event-list-item/event-list-item.scss
```diff
@@ -1,21 +1,64 @@
...
+  }
+  @include media-query__medium {
+    grid-template-columns:
+      $_event-list-item--image-width 1fr
+      max-content;
+    gap: $s-2xl;
+  }
+}
+
+@mixin event-content-layout {
```
LasseStaus, line 42:
```suggestion
@mixin event-list-item-content-grid {
```

src/stories/Library/event-list-item/event-list-item.scss
```diff
@@ -100,27 +141,52 @@ $_desktop-content-rows: 5;
...
+  @include event-layout;
+  @include event-list-item-stacked-shadow-box;
+  padding-top: $s-sm;
+  padding-bottom: $s-sm;
+  grid-template-areas: "contentStacked";
+  align-items: center;
+}
+
+.event-list-item-stacked__content {
+  @include typography($typo__body-placeholder);
```
LasseStaus, line 157:
```suggestion
  @include typography($typo__small-caption);
```

The way i see it, this font should be `$typo__small-caption`. 

The font sizes on the current figma for this, is not correct. The size on the stacked__content should be the same as the description on the eventlist item. This is using __small-caption. The reason the design for this file is using 14px for the description and not 12px, is because of a temporary change that is not accepted yet. 

`$typo__body-placeholder is not correct, since this is even higher (16px), i would advice we use the same as was used in the original description for event-list-item, `$typo__small-caption`

if this is used elsewhere, remember to change that there too.

src/stories/Library/event-list/event-list.scss
```diff
@@ -7,3 +7,7 @@
   list-style: none;
   margin-bottom: $s-md;
 }
+
+.event-list__item--stacked {
+  margin-top: -$s-md;
+}
```
LasseStaus, line 13:
```suggestion
.event-list__item + .event-list__item--stacked,
.event-list__item + .event-list__item {
  margin-bottm: $s-md;
}
```


Instead of using -margin, i would suggest to remove the margin-bottom from .event-list-item, and then use what i wrote above. I think it accomplishes the same. Personal preference of only "adding" margin instead of adding and then withdrawing it again.

src/stories/Library/event-list-item/event-list-item.scss
```diff
@@ -165,13 +231,19 @@ $_desktop-content-rows: 5;
     grid-row: 3 / span 3;
     justify-content: unset;
   }
+
+  .event-list-item-stacked {
+    grid-template-areas: ". contentStacked";
```
LasseStaus, line 236:
```suggestion
    grid-template-areas: ". content";
```

Again just suggestion: 

If everything is part of 'stacked' anyway, i would just call the areas for content, time, arrow, etc. If you prefer this that's also fine, up to you.

LasseStaus (approved):
It works responsively when i test it.
I have added some suggestions you can consider, but at minimum change the font. 😃

Will approve and let you do the changes. 

Nice one

src/stories/Library/event-list-item/event-list-item.scss
kasperbirch1: Do you know if it makes any difference to have several aria names with the same name? does it only look at the classes that are children of the grid?

PR #510, opened 2024-02-19
Fix nav-spot, to display placeholder box when no media. DDFFORM-309

rasben:
https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/779

### DIFF
--- a/src/stories/Library/nav-spot/NavSpot.tsx
+++ b/src/stories/Library/nav-spot/NavSpot.tsx
@@ -1,4 +1,5 @@
 import { FC, ReactNode } from "react";
+import clsx from "clsx";
 import { ReactComponent as Arrow } from "../Arrows/icon-arrow-ui/icon-arrow-ui-large-right.svg";
 
 type NavSpotProps = {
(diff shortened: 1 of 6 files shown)


kasperg (approved):
:+1: LGTM

src/stories/Library/nav-spot/NavSpot.tsx
```diff
@@ -1,21 +1,36 @@
...
       <a href="#" className="nav-spot__content">
-        {media ? <figure className="nav-spot__media">{media}</figure> : ""}
+        <figure className="nav-spot__media">
+          {media || (
+            <div className="nav-spot__placeholder-text">{placeholderText}</div>
+          )}
+        </figure>
+
         <div className="nav-spot__text">
           <h1 className="nav-spot__title">{title}</h1>
```
kasperg, line 35: It is outside the scope of this PR and heading levels are tricky but I would never expect a `<h1>` to be the right heading level for a Nav Spot.

kasperg (approved):
I had a few thoughts the second time around :smile:.

src/stories/Library/nav-spot/NavSpot.tsx
rasben: It's wrapped in a `<article>`, so yeah, h1 is appropiate :) @kasperg

src/stories/Library/nav-spot/NavSpot.tsx
```diff
@@ -1,21 +1,36 @@
...
+const NavSpot: FC<NavSpotProps> = ({
+  variant,
+  title,
+  subtitle,
+  media,
+  placeholderText,
+}) => {
+  const classes = clsx("nav-spot", "arrow__hover--right-large", {
+    "nav-spot--has-no-media": !media,
+    "nav-spot--has-media": media,
```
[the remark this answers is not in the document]
rasben: Good point - I dont use "has-media" - it is something that is used in cards.

PR #511, opened 2024-02-20
Fix various visual bugs

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-454

#### Description
Fixes various visual bugs:
- modal-loan content of modals wasn't spanning over full width of moals
- modal close button had a background color that was visible when the X was animating outside of modals

#### Screenshot of the result
-

-


spaceo (approved):
LGTM 👍

PR #514, opened 2024-02-20
Remove unnecessary SVG transform in event-list-item.scss

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-297

#### Description

There is enough space to keep the hover arrow in the same grid.


kasperg (approved):
:+1: LGTM

PR #517, opened 2024-02-21
DDFFORM 313

LasseStaus:
#### Link to issue

[DDFFORM-313](https://reload.atlassian.net/browse/DDFFORM-313)

- CMS PR: https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/796
- React PR: https://github.com/danskernesdigitalebibliotek/dpl-react/pull/971

#### Description

This PR add the design for Material Grid, showing a number of recommendedMaterials in a grid.  

#### Screenshot of the result

image

I am not fond of the placement of the title, compared to the width of the component. This should be the correct width's however.


src/stories/Library/material-grid/MaterialGrid.tsx
```diff
@@ -0,0 +1,50 @@
...
+  materials,
+}) => {
+  useEffect(() => {
+    require("./init-material-grid");
+  }, []);
+
+  const showMoreButton = materials.length > 8;
+
+  return (
+    <div className={clsx("material-grid")}>
```
kasperbirch1, line 26: Should this be a section for semantic reasons? and then I see no reason to use clsc when there is only one class?

src/stories/Library/material-grid/MaterialGrid.tsx
```diff
@@ -0,0 +1,50 @@
...
+      <div className="material-grid__title">{title}</div>
```
kasperbirch1, line 27: I also think this should be a h tag

src/stories/Library/material-grid/MaterialGrid.tsx
```diff
@@ -0,0 +1,50 @@
...
+      <div className="material-grid__items">
```
kasperbirch1, line 28: ul

src/stories/Library/material-grid/MaterialGrid.tsx
```diff
@@ -0,0 +1,50 @@
...
+        {materials.slice(0, amountOfMaterialsToShow).map((material, index) => (
+          <div
```
kasperbirch1, line 30: li

src/stories/Library/material-grid/init-material-grid.js
```diff
@@ -0,0 +1,20 @@
+window.addEventListener("load", () => {
```
kasperbirch1, line 1: I'm not sure if it's necessary but maybe we should use a DOMContentLoaded here

src/stories/Library/material-grid/init-material-grid.js
```diff
@@ -0,0 +1,20 @@
...
+  const materialGrids = document.querySelectorAll(
+    ".material-grid:not(.initialized)"
+  );
+
+  materialGrids.forEach((grid) => {
+    grid.classList.add("initialized");
+    const showMoreButton = grid.querySelector(".material-grid__show-more");
```
kasperbirch1, line 8: Is it better to capture elements with data attributes instead of class?

kasperbirch1 (approved):
I have some suggestions you could consider but or it looks fine to me 👍 

Remember to accept the UI changes in Chromatic before merging.

src/stories/Library/material-grid/MaterialGrid.tsx
LasseStaus: Used a <h2>

src/stories/Library/material-grid/init-material-grid.js
LasseStaus: I'm not sure if everything is loaded at that time once used in the drupal setup, but we can test it with DOMContent first and change it later if it's a problem. 👍

src/stories/Library/material-grid/MaterialGrid.tsx
LasseStaus: We add a section around it with the paragraph rendering in drupal. But yes 👍

src/stories/Library/material-grid/MaterialGrid.stories.tsx
```diff
@@ -0,0 +1,43 @@
...
+    },
+    href: {
+      control: "text",
+      description: "Link to the recommended material",
+    },
+    positionImageRight: {
+      control: "boolean",
+      description: "Toggle image position",
+    },
+  },
```
kasperg, line 25: Please reconsider these arguments.

Most of them are not relevant for this component. The only one that is - `title` has a somewhat misleading description.

src/stories/Library/material-grid/MaterialGrid.tsx
```diff
@@ -0,0 +1,51 @@
...
+};
+
+export const MaterialGrid: React.FC<MaterialGridProps> = ({
+  title,
+  amountOfMaterialsToShow,
+  materials,
+}) => {
+  // The JS code here is just an example of how this works.
+  // The proper implementation is in the react repo.
+  const [showAllMaterials, setShowAllMaterials] = useState(false);
```
kasperg, line 20: :+1: Nice commenting.

src/stories/Library/material-grid/MaterialGrid.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const [displayedItemCount, setDisplayedItemCount] = useState(8);
+  const showMoreButton = materials.length > 8;
```
kasperg, line 22: Consider your use of the value 8.

- Here the number 8 is used specifically. The ticket says that the default is 4. Figma shows 8 (but without a "Show more" button. I wonder what is right.

src/stories/Library/material-grid/MaterialGrid.tsx
```diff
@@ -0,0 +1,51 @@
...
+          <li key={index} className="material-grid__item">
+            <RecommendedMaterial {...material} />
+          </li>
+        ))}
+      </ul>
+      {showMoreButton && !showAllMaterials && (
+        <button
+          className="material-grid__show-more btn-primary btn-outline btn-medium"
+          data-show-more
+          aria-expanded="false"
```
kasperg, line 43: Please consider add an `aria-controls` attribute to the list.

:+1: remembering to add this `aria-expanded` attribute but [according to MDN you should also add an `aria-controls` attribute with the id of the element which is expanded when clicking the button](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded#buttons).

src/stories/Library/material-grid/MaterialGridData.tsx
```diff
@@ -0,0 +1,198 @@
+import { RecommendedMaterialProps } from "../recommended-material/RecommendedMaterial";
+
+const MaterialGridData: RecommendedMaterialProps[] = [
+  {
+    author: "Daniel Lee (2023)",
+    description: "Advanced Woodworking",
+    favoriteFill: false,
+    coverUrl: "#",
+    src: "images/book_cover_6.jpg",
+    alt: "book cover",
```
kasperg, line 10: Consider using an empty alt text. 

I see the cover images as descriptive. They provide no additional information compared to what is already available in the DOM. In these cases providing an empty `alt` text is the right thing to do.

src/stories/Library/material-grid/material-grid.scss
```diff
@@ -0,0 +1,30 @@
...
+}
+.material-grid__items {
+  display: grid;
+  grid-template-columns: repeat(2, 1fr);
+}
+
+.material-grid__items .recommended-material {
+  max-width: unset;
+  border: 1px solid $color__global-tertiary-1;
+}
```
kasperg, line 20: Please look into how we can avoid referencing another component.

According to our BEM guidelines one block should not modify the style of another block. Either we should look into if we can add modifiers to `.recommended-material` to make it behave like we want or we should create a `.material-grid__item` element within this block, style it appropriately and use this class on the `.recommended-material` elements.

kasperg (changes requested):
:+1: Nice but I think a few things needs a second look.

src/stories/Library/material-grid/MaterialGrid.stories.tsx
LasseStaus: Yea this is definitely incorrect, i shouldn't have missed this. 🙃

src/stories/Library/material-grid/MaterialGrid.tsx
LasseStaus: I understand. I've refactored the code and the naming of variables. 

From my understanding the initial max before showing a button should be 8, but i see they mentioned 4 in the ticket now. 

Change it to be using 4 now, and updated the figma link to the the specific component that fits description.

kasperg (approved):
:+1: LGTM

PR #518, opened 2024-02-22
Refactor and streamline the hero elements.

rasben:
Up until now, the same classes and components were being used almost identically, across several elements.
This PR creates a unified way of adding heros, both with and without links.

The easiest way to review this (apart from regular code review), is to look at chromatic changes:

https://www.chromatic.com/build?appId=616ffdab9acbf5003ad5fd2b&number=2150

There has been some spacing changes, that I feel match the design more as it is expected - I also introduce the "identity fallback" when there is no image available.


kasperbirch1:
There is something wrong with the button here.
https://www.chromatic.com/test?appId=616ffdab9acbf5003ad5fd2b&id=65d7546343a7364495b3067a

It does not follow the dialable design in Figma
https://www.figma.com/file/Zx9GrkFA3l4ISvyZD2q0Qi/Designsystem?type=design&node-id=7714-120937&mode=design&t=wAcqktP7xVkeVraT-4

rasben:
>Missing image on mobile?

We dont display the placeholder on mobile, as it doesnt really give any value there.

src/stories/Library/Heros/hero/HeroInner.tsx
```diff
@@ -0,0 +1,75 @@
...
+  date,
+  title,
+  description,
+  descriptionItems,
+  cta,
+  tag,
+}) => {
+  return (
+    <div className="hero__inner">
+      {tag ? (
```
kasperbirch1, line 28: Here you can use the && operator

Example:
function WelcomeMessage() {
  const [isLoggedIn, setIsLoggedIn] = React.useState(false);

  return (
    <div>
      {isLoggedIn && <h1>Velkommen tilbage!</h1>}
    </div>
  );
}

kasperbirch1 (approved):
LGTM 👍

PR #527, opened 2024-02-26
Make "EventListPage" more generic, to use it for non-events. DDFFORM-326

rasben:
Basically just renaming.

We need this list for the branch list (Fillial liste)


base.scss
```diff
@@ -143,7 +143,7 @@
 @import "./src/stories/Blocks/advanced-search/advanced-search";
 @import "./src/stories/Blocks/article/article";
 @import "./src/stories/Blocks/event-page/event-page";
-@import "./src/stories/Blocks/event-list-page/event-list-page";
+@import "./src/stories/Blocks/list-page/list-page";
```
kasperg, line 146: Please reconsider the name here.

I understand the intention but `list-page` is *very* generic. It would be nice if it was used across all lists like the search result and perhaps also other places, but that is not the case at the moment anyway.

How about something like `content-list-page`? `editorial-content-list-page` is quite lengthy...

src/stories/Library/event-list-item/event-list-item.scss
```diff
@@ -73,14 +73,14 @@ $_stacked-reduce-width: 20px;
...
-  object-fit: cover;
-  position: relative;
-  z-index: $z-10;
+  * {
+    width: 100%;
+    height: 100%;
+    object-fit: cover;
+    position: relative;
+    z-index: $z-10;
+  }
```
kasperg, line 83: Please consider if this change is needed.

I do not understand the need for this change. It does not seem to be related to the change at hand, it is not in a separate commit and it is not mentioned in the commit comment or in the pull request.

In general I would also think that a wildcard child selector would be something that we would want to avoid.

kasperg (changes requested):
:+1: I agree with the intent of this refactor and I think it follows the intention of the design with reusable list structures.

I have a couple of comments regarding naming and changes.

base.scss
rasben: I'll just go with `content-list-page`

src/stories/Library/event-list-item/event-list-item.scss
rasben: I get the confusion - I actually think this was meant to be in another PR

Either way - I've pulled it out into its own commit now, along with a code comment on why * is needed.

rasben:
Kasper: You had a good point about the `event-list-item`/`event-list` also needing new names, in the other PR.

I've added a new commit to this PR, that also generalizes that, so it's worth it for you to re-review everything again.

src/stories/Library/content-list-item/ContentListItemStacked.tsx
```diff
@@ -0,0 +1,29 @@
...
+  href = "#",
+  time,
+  date,
+}) => {
+  return (
+    <a
+      href={href}
+      className="content-list-item-stacked arrow__hover--right-small"
+    >
+      <time className="content-list-item-stacked__content">
```
kasperg, line 20: @kasperbirch1 has added a visually hidden title for the event in https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/785/commits/5d3dfe7db3bb945d6005a1a7393a1d8e78d776ff to improve accessibility.

Consider if this should be a part of the design system as well.

src/stories/Library/content-list-item/content-list-item.scss
```diff
@@ -67,23 +67,27 @@ $_stacked-reduce-width: 20px;
...
-  width: 100%;
-  height: 100%;
-  object-fit: cover;
-  position: relative;
-  z-index: $z-10;
+  // We want to treat everything inside the image-container as image related.
+  // To make object-fit work, it is important that everything has 100%
+  // height and width set. We cannot know 100% what markup Drupal will use,
+  // so we'll use a wildcard here.
+  * {
```
kasperg, line 81: It is hard for me to know what the exact problem is here.

The intention is for the design system to dictate the markup and for the consuming system to follow. I do not know what situation forces us to leave it as this.  Perhaps it would be easier to look at it on a screen together.

src/stories/Library/content-list/ContentList.tsx
```diff
@@ -0,0 +1,40 @@
...
+  items: ContentListItemProps[];
+};
+
+const ContentList: React.FC<ContentListProps> = ({ items }) => {
+  return (
+    <ul className="content-list">
+      {items.map((item, index) => {
+        // Check if current event should be stacked with the previous one
+        const isStacked =
+          items[index].eventSeriesId === items[index - 1]?.eventSeriesId ??
```
kasperg, line 18: This references something event specific in what is supposed to be a generic component.

I am not too happy about this. An alternate suggestion would be to just add a `stacked` attribute to each entry in the sample data and base the stacking on this.

kasperg (approved):
:+1: I appreciate the refactor and the additional work gone into this.

From a design system perspective I am not quite sure we nailed the abstraction between a generic list with generic items and a event list with event items that are stackable.

At least this is a start.

src/stories/Library/content-list-item/ContentListItem.tsx
```diff
@@ -30,32 +30,32 @@ export const EventListItem: React.FC<EventListItemProps> = ({
...
-        <h2 className="event-list-item__title">{title}</h2>
-        <div className="event-list-item__description">
+        <div className="content-list-item__date">{date}</div>
+        <h2 className="content-list-item__title">{title}</h2>
+        <div className="content-list-item__description">
           <p>{description}</p>
         </div>
-        <div className="event-list-item__location-wrapper">
-          <p className="event-list-item__location">{location}</p>
+        <div className="content-list-item__location-wrapper">
```
kasperbirch1, line 48: It's just an idea, but could you call content-list-item__bottom-wrapper and also have a content-list-item__right-wrapper If now the data inside changes?

PR #543, opened 2024-03-06
Ddfform 400 paragraph hero tilga link via billede

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-400
https://reload.atlassian.net/browse/DDFFORM-419

#### Description
This pull request includes a few minor CSS adjustments to the hero section. Please refer to the individual commits for a more detailed description of the changes

Its align with the changes in
https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/860

### DIFF
--- a/src/stories/Library/Heros/hero/HeroInner.tsx
+++ b/src/stories/Library/Heros/hero/HeroInner.tsx
@@ -1,5 +1,6 @@
 import React from "react";
 import Tag from "../../tag/Tag";
+import { ReactComponent as ArrowLargeRight } from "../../Arrows/icon-arrow-ui/icon-arrow-ui-large-right.svg";
 import ListDescription, {
   ListData,
 } from "../../Lists/list-description/ListDescription";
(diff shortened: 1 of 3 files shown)


src/stories/Library/Heros/hero/HeroInner.tsx
```diff
@@ -44,6 +47,12 @@ const HeroInner: React.FunctionComponent<HeroInnerProps> = ({
       {description && (
         <div className="hero__description">
           <p>{description} </p>
+
+          {arrow && (
+            <div>
```
kasperg, line 52: I am not a fan of these filler seemingly unnecessary divs.

As I understand it it is needed because we want to apply `margin-bottom` to `.hero__description > *` and we need a block element to do that. The arrow element is not a block. 

This is totally outside the scope of this PR though.

kasperg (approved):
:+1: LGTM

PR #547, opened 2024-03-11
Refactor link styling in `ListDescription` component

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-404

#### Description

This pull request refactors the link styling in the `ListDescription` component. It adds correct link markup and styles the link according to the Drupal rendering method.

https://github.com/danskernesdigitalebibliotek/dpl-react/blob/develop/src/components/material/MaterialDetailsList.tsx
I have reviewed the `ListDescription` component in dpl-react and noticed that it currently does not render any links. Therefore, this will need to be revisited when we implement logic to display links in the `ListDescription` in the future.

### DIFF
--- a/src/stories/Library/Lists/list-description/list-description.scss
+++ b/src/stories/Library/Lists/list-description/list-description.scss
@@ -27,7 +27,11 @@ dl.list-description {
     flex-direction: column;
   }
 
-  .link-tag {
+  // TODO: Reevaluate this when implementing logic to determine if 'link-tag' styled
+  // values in dpl-react should be 'a' tags or 'span' tags as currently implemented.
+  .link-tag,
+  a {
+    @extend %link-tag;
     @include typography($typo__small-caption);
 
     color: $color__text-primary-black;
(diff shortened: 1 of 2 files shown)


src/stories/Library/Lists/list-description/list-description.scss
```diff
@@ -27,7 +27,11 @@ dl.list-description {
     flex-direction: column;
   }
 
-  .link-tag {
+  // TODO: Reevaluate this when implementing logic to determine if 'link-tag' styled
+  // values in dpl-react should be 'a' tags or 'span' tags as currently implemented.
+  .link-tag,
```
kasperg, line 32: As I see it we should drop using *and* supporting `link-tag` for description elements that are not links.

Using nested elements is against our BEM structure and it is just odd to have something that looks like an interactive element but is not.

I think you need to clear this with the LSBP team before proceeding.

kasperg (changes requested):

Adamik10 (commented):
I tried looking through the links, but I'm still not sure exactly which element we're talking about from the connected Figma screen (link in the ticket).
But I agree with @kasperg that if the element isn't interactive, it shouldn't give off that impression to the end user.

kasperbirch1:
@Adamik10 
> I tried looking through the links, but I'm still not sure exactly which element we're talking about from the connected Figma screen (link in the ticket). But I agree with @kasperg that if the element isn't interactive, it shouldn't give off that impression to the end user.

Its in the material details 
https://www.figma.com/file/Zx9GrkFA3l4ISvyZD2q0Qi/Designsystem?type=design&node-id=8513-62957&mode=design&t=cYSDsVc2dSgBTcoc-4

kasperg:
@kasperbirch1 I am reassigning to you to decide how to move this forward.

PR #550, opened 2024-03-14
Move modal close button

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-500

#### Description
Move modal close button 17px to the left so scrollbar isn't obscured - 17 because that is the width of the scrollbar on Windows, which is the thickest one.

#### Screenshot of the result
-


Dresse (approved):
LGTM - However the base branch needs to be updated before merging :)

PR #559, opened 2024-03-18
Event facets

kasperg:
#### Link to issue

https://reload.atlassian.net/browse/DDFFORM-65

#### Description

This updates the design of content pages to support multiple facets. This includes visual separation and handling across mobile and desktop.

Please see the individual commits for additional commentary.

#### Screenshot of the result

See Chromatic.

For some reason Chromatic lists a number of unrelated pages as having changed without there being any actual changes to spot. I have accepted these changes as well.


LasseStaus (approved):
LGTM 👍 🚀 

I notice the width of the individual dropdown seems to be determined by the length of the longest option. Consider if this the behaviour we want, or if there should be a max-width before it gets truncated.

PR #567, opened 2024-04-03
Renewal reason should fit in on cards

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-405

#### Description
This PR plays around with selectable card status message placement.

#### Screenshot of the result
Before:
image

After:
image

After on mobile:
image

This PR will have a sibling in dpl-react when this is merged in.


Dresse (approved):
I think this looks good, although I would suggest adding a little padding between the title of the material and the author. Right now it feels like it is almost overlapping. 

I'll approve the PR however!

Adamik10:
I didn't fiddle around with that spacing at all 😳

PR #568, opened 2024-04-03
DDFFORM 417

LasseStaus:
#### Link to issue

Jira: [DDFFORM-417](https://reload.atlassian.net/browse/DDFFORM-417)

#### Description

This PR depends on two PRs. 

[React PR](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/1076)
[CMS PR](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/937)

This PR adds the markup and styling for the OpeningHours component, which is used to display the opening hours for a branch. 

I have left chromatic unaccepted on purpose. Will accept once approved. 

#### Screenshot of the result

image

I am not sure why there are  34 UI changes. 

The only changes that _should_ be related to this pr, are the New Stories for openingHours in chromatic. 

Please add comments if you have any suggestions for improvement. 

This PR had a lot of refactoring after developing the React part. If you see any unused CSS or unnessecary markup mark it, although this hopefully isn't the case. 

[DDFFORM-417]: https://reload.atlassian.net/browse/DDFFORM-417?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/stories/Library/opening-hours/OpeningHours.tsx
+++ b/src/stories/Library/opening-hours/OpeningHours.tsx
@@ -1,0 +1,46 @@
+import { FC } from "react";
+import { ReactComponent as ArrowLeft } from "../../../public/icons/collection/ArrowLeft.svg";
+import { ReactComponent as ArrowRight } from "../../../public/icons/collection/ArrowRight.svg";
+import { GroupedOpeningHours } from "./OpeningHoursExampleData";
+import OpeningHoursWeekList from "./OpeningHoursWeekList";
+
+type OpeningHoursProps = {
+  title: string;
+  weekCurrentlyDisplayed: string;
+  groupedOpeningHours: GroupedOpeningHours;
+};
+
+const OpeningHours: FC<OpeningHoursProps> = ({
+  groupedOpeningHours,
+  weekCurrentlyDisplayed,
+}) => {
+  return (
+    <div className="opening-hours">
+      <div className="opening-hours__header">
+        <h2 className="opening-hours__heading">Åbningstider</h2>
+        <div className="opening-hours__navigation-controls">
+          <button
+            className="opening-hours__navigation-control"
+            aria-label="Previous week"
+            type="button"
+          >
+            <ArrowLeft />
+          </button>
+          <div className="opening-hours__week-display">
+            {weekCurrentlyDisplayed}
+          </div>
+          <button
+            className="opening-hours__navigation-control"
+            aria-label="Next week"
+            type="button"
+          >
+            <ArrowRight />
+          </button>
+        </div>
+      </div>
+      <OpeningHoursWeekList groupedOpeningHours={groupedOpeningHours} />
+    </div>
+  );
+};
+
+export default OpeningHours;
--- a/src/stories/Library/opening-hours/OpeningHoursExampleData.ts
+++ b/src/stories/Library/opening-hours/OpeningHoursExampleData.ts
@@ -1,0 +1,250 @@
+// These types are based on the actual data structure returned from the API.
+// The types starting with DplOpeningHours.. are generated with Orval in react
+// and copy pasted here.
+export type DplOpeningHoursListGET200Item = {
+  /** An serial unique id of the opening hours instance. */
+  id: number;
+  category: DplOpeningHoursListGET200ItemCategory;
+  /** The date which the opening hours applies to. In ISO 8601 format. */
+  date: string;
+  /** When the opening hours start. In format HH:MM */
+  start_time: string;
+  /** When the opening hours end. In format HH:MM */
+  end_time: string;
+  /** The id for the branch the instance belongs to */
+  branch_id: number;
+};
+export type DplOpeningHoursListGET200ItemCategory = {
+  title: string;
+  /** A CSS compatible color code which can be used to represent the category */
+  color: string;
+};
+
+export type GroupedOpeningHours = Array<{
+  day: string;
+  date: string;
+  openingHourEntries: DplOpeningHoursListGET200Item[];
+}>;
+
+// Example data in the format it is used in the grouped opening hours component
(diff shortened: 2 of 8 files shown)


src/stories/Library/opening-hours/OpeningHoursExampleData.ts
```diff
@@ -0,0 +1,250 @@
+// These types are based on the actual data structure returned from the API.
+// The types starting with DplOpeningHours.. are generated with Orval in react
+// and copy pasted here.
+export type DplOpeningHoursListGET200Item = {
```
kasperg, line 4: :+1: Appreciate the commenting here.

Normally I would not think that these "technicalities" belonged in the design system"

src/stories/Library/opening-hours/OpeningHours.tsx
```diff
@@ -0,0 +1,48 @@
...
+              aria-label="Next week"
+              type="button"
+            >
+              <ArrowRight />
+            </button>
+          </div>
+        </div>
+        <OpeningHoursWeekList groupedOpeningHours={groupedOpeningHours} />
+      </div>
+    </div>
```
kasperg, line 44: Consider if we can drop the `div className="opening-hours__wrapper">` element.

We already have a `<div className="opening-hours">` element to wrap the entire component.

src/stories/Library/opening-hours/OpeningHoursWeekList.tsx
```diff
@@ -0,0 +1,48 @@
...
+            day: "numeric",
+            month: "2-digit",
+          }
+        );
+
+        return (
+          <li key={date} className="opening-hours__row">
+            <h3 className="opening-hours__individual-day">{`${day}: d. ${formattedDateForDisplay}`}</h3>
+            {openingHourEntries.length > 0 ? (
+              <ul>
```
kasperg, line 27: Consider if these lists are actually ordered (`<ol>` instead of `<ul>`).

In general I appreciate the use of list elements here.

src/stories/Library/opening-hours/OpeningHoursWeekList.tsx
```diff
@@ -0,0 +1,48 @@
...
+          "da-DK",
+          {
+            day: "numeric",
+            month: "2-digit",
+          }
+        );
+
+        return (
+          <li key={date} className="opening-hours__row">
+            <h3 className="opening-hours__individual-day">{`${day}: d. ${formattedDateForDisplay}`}</h3>
```
kasperg, line 25: Consider using a [`<time>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time) here.

src/stories/Library/opening-hours/opening-hours.scss
```diff
@@ -0,0 +1,89 @@
...
+
+.opening-hours__individual-opening {
+  display: flex;
+  justify-content: space-between;
+  padding: 0 $s-xs;
+  @include media-query__small {
+    line-height: 32px;
+  }
+}
+.opening-hours__individual-opening--odd {
```
kasperg, line 55:
```suggestion
.opening-hours__individual-opening:nth-child(odd)  {
```

I think this can be simplified a bit using [`:nth-child()`](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child#trnth-childodd) like we use `:first-child` below.

This means we have remove the logic for applying this class also.

src/stories/Library/opening-hours/opening-hours-skeleton.scss
```diff
@@ -0,0 +1,13 @@
+.opening-hours {
+  .ssc-line {
+    height: $s-xl;
+  }
+  .ssc-line--odd {
+    opacity: 0.5;
+  }
```
kasperg, line 7: :+1: for making an effort by adding a skeleton.

src/stories/Library/opening-hours/opening-hours.scss
```diff
@@ -0,0 +1,89 @@
...
+  height: $_nav-control-height;
+  background-color: $color__global-primary;
+  border: $_nav-control-border-width solid $color__global-tertiary-1;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+
+  &:hover {
+    cursor: pointer;
+  }
```
kasperg, line 79: Consider if this is necessary as these element are `<buttons>` already.

kasperg (approved):
:+1: I have a few things for you to consider but you can decide for yourself how/if to solve these and then merge at you please.

src/stories/Library/opening-hours/OpeningHoursWeekList.tsx
LasseStaus: I think it makes sense to used OL since they atleast the weeks follow an specific sequence. 👍  Will for "categories" aswell.

src/stories/Library/opening-hours/OpeningHours.tsx
LasseStaus: I need the layout-container for opening_hours, but the contents need to be max-width:600px. within that. 

I wasn't sure whether to use a wrapper for all the content, or to add the max-width two places, in this case __header and __content. 

I have changed it to the second approach now, which effectively does the same, but without the wrapper. Not sure which i like best.

src/stories/Library/opening-hours/OpeningHoursWeekList.tsx
LasseStaus: Yeah agree 👍

src/stories/Library/opening-hours/opening-hours.scss
LasseStaus: Yeah this is better 👍

PR #578, opened 2024-04-08
Super simple blockquote design, for rich-text. DDFFORM-492

rasben:
There is no design on this element, but if we add nothing, then there is no visual indication that this is a blockquote. This just adds a very simple, visual indicator.

https://reload.atlassian.net/browse/DDFFORM-492

Screenshot 2024-04-08 at 16 38 51

### DIFF
--- a/src/stories/Library/rich-text/rich-text.scss
+++ b/src/stories/Library/rich-text/rich-text.scss
@@ -75,6 +75,25 @@ $_max-width-rich-text: $block-max-width__small;
     }
   }
 
+  blockquote {
+    $_quote-indicator-width: 10px;
+
+    position: relative;
+    padding-top: $s-md;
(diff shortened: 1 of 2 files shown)


src/stories/Library/rich-text/rich-text.scss
```diff
@@ -75,6 +75,25 @@ $_max-width-rich-text: $block-max-width__small;
     }
   }
 
+  blockquote {
+    $_quote-indicator-width: 10px;
```
kasperg, line 79: :+1: I liked this use of a variable.

I wonder if there is a reason why you went with `10px` instead of e.g. `8px` (`$s-sm`) which is already a size we use in the design.

kasperg (approved):
:+1: LGTM

PR #582, opened 2024-04-10
Add Patron page and skeleton [DDFLSBP-486]

clausbruun:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-486

#### Description

Based on the version of the Patron page from the dpl-react project, I have constructed a Patron page with an accompanying skeleton. Some decisions were made to reuse parts of dpl-react where immediately possible, but also to use components from the design system where they already existed, instead of the version that emerged in dpl-react.

#### Screenshot of the result

Skærmbillede 2024-04-10 kl  13 00 50

*

### DIFF
--- a/src/stories/Blocks/patron-page/ContactInfoSection.tsx
+++ b/src/stories/Blocks/patron-page/ContactInfoSection.tsx
@@ -1,0 +1,30 @@
+import React from "react";
+import Input from "../../Library/Forms/input/Input";
+import { Checkbox } from "../../Library/Forms/checkbox/Checkbox";
+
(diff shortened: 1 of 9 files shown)


Adamik10 (dismissed):
Very nice! 
I would really like to see some screenshots in the PR description, since we are in the design system 🙈

Adamik10 (approved):
Very nice :) good job.

clausbruun:
> Very nice :) good job.

Thanks! I just remembered that I hadn't updated the description, now done :D

PR #583, opened 2024-04-10
Fix iOS 15 and later blue button text color [DDFLSBP-504]

clausbruun:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-504

#### Description

Corrects the issue where, in iOS 15 and later, if no color is specified in CSS for the button element, the text color defaults to blue. As a side effect, buttons defined in Figma to have the color #484848 now display the correct color in other browsers as well, where previously they defaulted to #000, the standard when no color is specified.

Fixed for:
- Footer
- Advanced Search
- Search Result Page
- Buttons / RowButton
- Lists / ReservationsAndLoans
- Modals / Facet browser
- Tag / Tag Button

### DIFF
--- a/src/stories/Library/Modals/modal-find-on-shelf/ModalFindOnShelf.tsx
+++ b/src/stories/Library/Modals/modal-find-on-shelf/ModalFindOnShelf.tsx
@@ -66,7 +66,7 @@ const ModalFindOnShelf: React.FC<ModalFindOnShelfProps> = ({
             headingLevel="h2"
             withAvailability
             fullWidth
-            headline="Bibliotek fliale navn"
+            headline="Bibliotek filial navn"
             icon="Various"
             key={branchKey}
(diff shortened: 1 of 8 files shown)


src/stories/Library/Modals/modal-find-on-shelf/ModalFindOnShelf.tsx
```diff
@@ -66,7 +66,7 @@ const ModalFindOnShelf: React.FC<ModalFindOnShelfProps> = ({
             headingLevel="h2"
             withAvailability
             fullWidth
-            headline="Bibliotek fliale navn"
+            headline="Bibliotek filial navn"
```
Adamik10, line 69: Thx 🙏🏼

Adamik10 (approved):
Yay!

PR #587, opened 2024-04-12
Make sure accordion body has rich-text/WYSIWYG styling. DDFFORM-467

rasben:
Related to 

https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/988

https://reload.atlassian.net/browse/DDFFORM-467

image


rasben:
The chromatic changes are unrelated - It's part of the bug, that we're looking into. I'll leave them as yellow.

Dresse (approved):
LGTM

PR #589, opened 2024-04-12
Fix inteligent facets on mobile devices

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-518

#### Description
This PR makes the filter facet line horizontally scrollable on mobile in one line in order to save space on smaller devices.

#### Screenshot of the result
Mobile:

https://github.com/danskernesdigitalebibliotek/dpl-design-system/assets/28546954/267455f3-788e-4910-91f1-029c8629225d

Desktop (stays the same):
image

-

### DIFF
--- a/src/stories/Library/card-list-page/facet-line.scss
+++ b/src/stories/Library/card-list-page/facet-line.scss
@@ -34,7 +34,7 @@
   white-space: nowrap;
 
   @include media-query__small() {
-    margin-bottom: $s-sm;
+    margin-top: $s-sm;
   }
 
   .tag {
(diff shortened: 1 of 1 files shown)


spaceo (approved):
Nice solution!! 👏 
I I am just wondering about two thing:
Is it clear to the user that the facet band is scrollable?
Is there an obvious ux trick to emphasize that?
Have you tested it on both Ios and Android?

Anyways, I'll approve it.
👍 LGTM

Adamik10:
Just to reply to the comment:

- I spoke with Tue about it and we found an example in the GoMore app - so we are trusting the UX designers on their end.
- No, Only tested on my PC, actually, as I can't run the code locally on any phone. It shall be tested on the test site.

PR #601, opened 2024-04-19
Fix broken reverts

rasben:
I previously had to revert commits, because i had merged them into develop by accident.
Something went completely wrong when develop was merged back into hermod-sprint-12, so they got removed again.

Long story short ^

Long story long: V

https://reload.zulipchat.com/#narrow/stream/240325-DDF/topic/develop.20troubles


rasben:
Related:

https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1019

kasperg (approved):

PR #607, opened 2024-04-23
DDFLSBP-528/add status message story and styling for succesful webform submission

Dresse:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-528

#### Description

When the user submits a webform, the submission notification status message was not styled. Thus I created a new story for status-messages. It largely resembles the error-message styling, but using the icon-check instead of icon-warning.

#### Screenshot of the result

Design system: short message:
Screenshot 2024-04-24 at 00 03 31

Design system: long message:
Screenshot 2024-04-24 at 00 04 44

### DIFF
--- a/src/stories/Library/status-message/StatusMessage.stories.tsx
+++ b/src/stories/Library/status-message/StatusMessage.stories.tsx
@@ -1,0 +1,26 @@
+import { ComponentMeta, ComponentStory } from "@storybook/react";
+import { withDesign } from "storybook-addon-designs";
+
+import { StatusMessage } from "./StatusMessage";
+
+export default {
+  title: "Library / Status Message",
+  component: StatusMessage,
+  decorators: [withDesign],
+  argTypes: {},
+  parameters: {},
+} as ComponentMeta<typeof StatusMessage>;
+
+const Template: ComponentStory<typeof StatusMessage> = (args) => (
+  <StatusMessage {...args} />
+);
+
+export const ShortStatus = Template.bind({});
+ShortStatus.args = {
+  message: "Handling gennemført.",
+};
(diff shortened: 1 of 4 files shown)


src/stories/Library/status-message/StatusMessage.stories.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+const Template: ComponentStory<typeof StatusMessage> = (args) => (
+  <StatusMessage {...args} />
+);
+
+export const ShortStatus = Template.bind({});
+ShortStatus.args = {
+  message: "Handling gennemført.",
+};
+export const LongStatus = Template.bind({});
```
Adamik10, line 22: Very good to show multiple states it can be in 👏

src/stories/Library/status-message/status-message.scss
```diff
@@ -0,0 +1,60 @@
...
+  &__icon {
+    margin-right: $s-sm;
+    min-width: 40px;
+  }
+
+  &__description {
+    color: $color__text-primary-black;
+    @include typography($typo__body-placeholder);
+    padding: $s-md $s-sm;
+    max-width: 82%;
```
Adamik10, line 22: is there a reason why the description can only be max 82% wide?

Adamik10 (approved):
Very good job! I got a question, but nothing blocking 😉

src/stories/Library/status-message/status-message.scss
Dresse: To be honest I have done a copy/paste of the error-message.scss, so I don't really know why it is like that. 😅

PR #609, opened 2024-04-26
Fix/mobile sidebar accessible on desktop

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-555

#### Description
- Make the header-sidebar-nav not display on desktop. Otherwise it's accessible by keyboard, which is an accessibility issue.
- Make sure the "Header sidebar nav" story displays on desktop. Despite the fact that the sidebar nav isn't supposed to be displayed at all on desktops. We just make sure to display it in storybook.

#### Screenshot of the result
-
-


Dresse (approved):
Maybe this is an obvious question, but why do we wan't to see it in Storybook? 

I'll approve this, as it looks good otherwise 😄

PR #610, opened 2024-04-26
Added list type to ListDescription values [DDFLSBP-543]

clausbruun:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-543

#### Description

Added 'list' type to ListDescription for rendering content descriptions, such as track listings.

#### Screenshot of the result

image

### DIFF
--- a/src/stories/Library/Lists/list-description/list-description.scss
+++ b/src/stories/Library/Lists/list-description/list-description.scss
@@ -26,6 +26,11 @@ dl.list-description {
   .list-description__value--column {
     flex-direction: column;
   }
+  
+  .list-description__value--list {
+    list-style: initial;
+    padding-left: 1em;
+  }
 
   .link-tag {
     @include typography($typo__small-caption);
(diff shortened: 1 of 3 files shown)


src/stories/Library/Lists/list-description/list-description.scss
```diff
@@ -27,6 +27,11 @@ dl.list-description {
     flex-direction: column;
   }
 
+  .list-description__value--list {
+    list-style: initial;
+    padding-left: 1em;
```
Adamik10, line 32: Maybe I would like to see some of our pre-defined spacing variables here instead, as we use those all around (check out the variables.spacing file).

Adamik10 (approved):
Approved! Although I do have a comment I'd like you to check out 😉

src/stories/Library/Lists/list-description/list-description.scss
clausbruun: Thank you, @Adamik10. I have updated it to`padding-left: $s-md;`. This translates to `16px` (where `1em` is approximately `12px`), and I actually think it looks better now. 👍🏻

image

PR #614, opened 2024-04-29
Encode inline SVGs

kasperg:
#### Description

When using an inline SVG it must be encoded. Otherwise browsers may choke on it due to unescaped characters like < and ".

At the moment problem can be seen when the Chromatic/Cypress integration tries to generate pages with inline CSS. Without this process fails and CSS is never loaded.

To address this we encode the inline CSS using
https://yoksel.github.io/url-encoder/. A proper SCSS function to address this would be preferable but I could not find such a thing.

We only have a few actual instances of inline SVGs to for now it should be OK to just change these.

This is needed for danskernesdigitalebibliotek/dpl-cms#1333

### DIFF
--- a/src/stories/Library/material-header/material-periodical-select.scss
+++ b/src/stories/Library/material-header/material-periodical-select.scss
@@ -20,7 +20,7 @@ $icon-position: 10px;
       background-repeat: no-repeat;
       background-size: 6px;
       // This is a copy from  icons/basic/icon-select-arrow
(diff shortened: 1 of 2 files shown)


kasperg:
The Chromatic tests currently fail for an SVG. We need to fix this before this can be merged.

kasperg:
> The Chromatic tests currently fail for an SVG. We need to fix this before this can be merged.

This has now been fixed.

rasben (approved):

PR #616, opened 2024-05-01
User registration form v2

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-563
https://reload.atlassian.net/browse/DDFLSBP-594

#### Description
This PR introduces the Create Patron story, and updates some styling connected to form elements.

#### Screenshot of the result
image
image

[Sibling PR](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/1150)

### DIFF
--- a/src/stories/Library/Forms/input/DoubleInputRow.tsx
+++ b/src/stories/Library/Forms/input/DoubleInputRow.tsx
@@ -3,23 +3,35 @@ import Input from "./Input";
 export type DoubleInputRowProps = {
   leftLabel: string;
   rightLabel: string;
+  descriptionLeft?: string;
+  descriptionRight?: string;
+  validationLeft?: string;
+  validationRight?: string;
 };
 
 const DoubleInputRow: React.FC<DoubleInputRowProps> = ({
   leftLabel,
   rightLabel,
+  descriptionLeft,
+  descriptionRight,
+  validationLeft,
+  validationRight,
 }) => {
   return (
     <div className="dpl-input__flex">
       <Input
         id={leftLabel}
         label={leftLabel}
+        description={descriptionLeft}
+        validation={validationLeft}
         type="text"
         classNames="dpl-input dpl-input--double mr-16"
       />
       <Input
         id={rightLabel}
         label={rightLabel}
+        description={descriptionRight}
+        validation={validationRight}
         type="text"
         classNames="dpl-input dpl-input--double"
       />
--- a/src/stories/Library/Forms/input/DoubleInputRow.tsx
+++ b/src/stories/Library/Forms/input/DoubleInputRow.tsx
@@ -1,0 +1,30 @@
+import Input from "./Input";
+
+export type DoubleInputRowProps = {
+  leftLabel: string;
+  rightLabel: string;
+};
+
+const DoubleInputRow: React.FC<DoubleInputRowProps> = ({
+  leftLabel,
+  rightLabel,
+}) => {
+  return (
+    <div className="contact-info-flex">
+      <Input
+        id={leftLabel}
+        label={leftLabel}
+        type="text"
+        classNames="patron__input patron__input--desktop mr-16"
+      />
+      <Input
+        id={rightLabel}
+        label={rightLabel}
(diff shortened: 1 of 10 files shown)


src/stories/Library/Forms/input/DoubleInputRow.tsx
```diff
@@ -0,0 +1,30 @@
+import Input from "./Input";
+
+export type DoubleInputRowProps = {
+  leftLabel: string;
+  rightLabel: string;
+};
+
+const DoubleInputRow: React.FC<DoubleInputRowProps> = ({
```
spaceo, line 8: Nice addition!

src/stories/Library/Forms/input/DoubleInputRow.tsx
```diff
@@ -0,0 +1,30 @@
...
+  leftLabel,
+  rightLabel,
+}) => {
+  return (
+    <div className="contact-info-flex">
```
spaceo, line 13: The naming and placemnt of this component indicates that it is a general form input component. But this class name is very specific. What are youre thoughts here

src/stories/Library/Forms/input/DoubleInputRow.tsx
```diff
@@ -0,0 +1,30 @@
...
+        id={leftLabel}
+        label={leftLabel}
+        type="text"
+        classNames="patron__input patron__input--desktop mr-16"
+      />
+      <Input
+        id={rightLabel}
+        label={rightLabel}
+        type="text"
+        classNames="patron__input patron__input--desktop"
```
spaceo, line 24: Same comment as the previous "patron__input" is specifc compare to the palcement/naming of the `DoubleInputRow` component

src/stories/Library/Forms/input/Input.tsx
```diff
@@ -6,13 +6,14 @@ export type InputProps = {
...
+  classNames?: string;
 };
 
 const Input = (props: InputProps) => {
-  const { label, type, id, description, validation } = props;
+  const { label, type, id, description, validation, classNames } = props;
   const invalid = validation ? "true" : "false";
   return (
-    <div className="dpl-input">
+    <div className={`dpl-input ${classNames || ""}`}>
```
spaceo, line 16: Why not using clsx ? :)

spaceo (changes requested):
Again a well delivered piece of work by mr Antal.
Have some comments though...

src/stories/Library/Forms/input/DoubleInputRow.tsx
Adamik10: Yesss, it's because we already have that class and use it in the V1 component. I could make a copy of it though, with the same properties if you think that works better :)

src/stories/Library/Forms/input/DoubleInputRow.tsx
Adamik10: And the answer is also same as above.

src/stories/Library/Forms/input/Input.tsx
```diff
@@ -6,13 +7,14 @@ export type InputProps = {
...
+  classNames?: string;
 };
 
 const Input = (props: InputProps) => {
-  const { label, type, id, description, validation } = props;
+  const { label, type, id, description, validation, classNames } = props;
   const invalid = validation ? "true" : "false";
   return (
-    <div className="dpl-input">
+    <div className={clsx("dpl-input", !!classNames && classNames)}>
```
clausbruun, line 17:
```suggestion
    <div className={clsx("dpl-input", classNames)}>
```
I think `clsx()` understands falsy values?

src/stories/Library/Forms/input/DoubleInputRow.tsx
spaceo: I think we should go as generic as possible moving forward. So if it implies copying a component(s) and consider the old ones deprecated, I think we should do that. We can even write in the old ones that they are deprecated -> even name them `BlaBlaDeprecated.tsx`

PR #619, opened 2024-05-02
Add design for the registration redirect login message

spaceo:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-615

#### Description

Adds a story showing the redirect login message after submission of the patron registration form.

##### Screenshots
image

 Belongs to a trinity together with:
+ https://github.com/danskernesdigitalebibliotek/dpl-react/pull/1166
+ https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1091


base.scss
```diff
@@ -163,6 +163,7 @@
 @import "./src/stories/Blocks/reservation-page/reservation-page-skeleton";
 @import "./src/stories/Blocks/loan-page/loan-page-skeleton";
 @import "./src/stories/Blocks/registration-page/registration-page";
+@import "./src/stories/Blocks/registration-formular-page/registration-formular-page";
```
Dresse, line 166: Maybe this should just be `registration-form-page/registration-form-page`? I don't think "formular" is an english word. It would either have to be "form" or "formula", but the latter is also used in other contexts.

src/stories/Blocks/registration-formular-page/RegistrationFormularPage.stories.tsx
```diff
@@ -0,0 +1,31 @@
+import { ComponentStory } from "@storybook/react";
+import RedirectToLoginMessage, {
+  RedirectToLoginMessageProps,
+} from "./RedirectToLoginMessage";
+
+export default {
+  title: "Blocks / Registration Formular Page",
```
Dresse, line 7: Same as above.

Dresse (approved):
LGTM - Have one comment regarding using the word "form" instead of "formular".

base.scss
spaceo: 👍

spaceo:
Moved to separate release:
https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/625

PR #622, opened 2024-05-03
Refactor `content-list-item__description`

kasperbirch1:
#### Depend on:
https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1055

#### Description
Remove unnecessary `p` tag and set margin back to default `auto`, which is needed for the truncate to work properly.


LasseStaus (approved):
LGTM! 🚀

PR #626, opened 2024-05-08
Fix/material item appearance

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-560

#### Description
Remove disclosure dropdown arrow in Safari + Remove interactivity effect from editions card items on material page. These aren't ever clickable so it doesn't make sense and could be misleading to the user.

#### Screenshot of the result
image

-


spaceo (approved):
Niceissity 👍

PR #631, opened 2024-05-15
Add focus on next item after "show more" click in `nav-grid`

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-612

#### Description
This pull request ensures that the next item link in the `nav-grid` receives focus after the "show more" button is clicked.

#### Screenshot of the result

https://github.com/danskernesdigitalebibliotek/dpl-design-system/assets/49920322/847b75e1-ac7c-4335-b7be-2209082ade20

### DIFF
--- a/src/stories/Library/nav-grid/init-nav-grid.js
+++ b/src/stories/Library/nav-grid/init-nav-grid.js
@@ -13,6 +13,15 @@ window.addEventListener("load", () => {
 
     button.addEventListener("click", () => {
       grid.classList.remove("nav-grid--folded");
+
+      // The 7th item is the first one that is initially hidden.
+      // For more details, refer to the ".nav-grid--folded" class in design-system src/stories/Library/nav-grid/nav-grid.scss
+      const seventhNavGridItem = grid.querySelector(
+        ".nav-grid__item:nth-child(7)"
+      );
+      if (seventhNavGridItem) {
+        seventhNavGridItem.querySelector("a").focus();
+      }
     });
   });
 });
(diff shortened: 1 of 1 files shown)


src/stories/Library/nav-grid/init-nav-grid.js
```diff
@@ -13,6 +13,15 @@ window.addEventListener("load", () => {
 
     button.addEventListener("click", () => {
       grid.classList.remove("nav-grid--folded");
+
+      // The 7th item is the first one that is initially hidden.
+      // For more details, refer to the ".nav-grid--folded" class in design-system src/stories/Library/nav-grid/nav-grid.scss
+      const seventhNavGridItem = grid.querySelector(
+        ".nav-grid__item:nth-child(7)"
```
rasben, line 20: This is waaay too magical.
Instead of looking for the 7th, you should instead try to find the first one that is hidden or something :)

rasben (changes requested):

src/stories/Library/nav-grid/init-nav-grid.js
kasperbirch1: How can I do it when the display none logic is not set on the items but on the container?

```
.nav-grid--folded {
  .nav-grid__item:nth-child(n + 7) {
    display: none;
  }

rasben: You could either check the CSS:

https://stackoverflow.com/questions/4866229/check-element-css-display-with-javascript

Or, check if it is visible:

https://stackoverflow.com/a/21696585

kasperbirch1: Thanks 🙏

rasben (approved):

PR #636, opened 2024-05-23
DDFFORM 611 backend anbefalings komp

LasseStaus:
#### Link to issue

Ticket: [DDFFORM-611](https://reload.atlassian.net/browse/DDFFORM-611)

This PR Depends on: 

React PR: https://github.com/danskernesdigitalebibliotek/dpl-react/pull/1223
CMS PR: https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1169

#### Description

This PR introduces the markup and styling for MaterialSearch, a component that will be used to allow searching for materials and selecting them for display in the admin interface for paragraphs using the the work_id field with the new field widget. Currently this is Recommandation and MaterialGridManual. 

You can either go commit by commit or view it as a whole. 

Consider the introduction of admin-base css and comment if there are any relevant details i have missed. 

Consider both the comment and approach in the css file regarding covers. 

Remember to thoroughly test this in drupal with all changes in cohesion.

#### Screenshot of the result

image

[DDFFORM-611]: https://reload.atlassian.net/browse/DDFFORM-611?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/base.scss
+++ b/base.scss
@@ -142,6 +142,7 @@
 @import "./src/stories/Library/opening-hours/opening-hours-skeleton";
 @import "./src/stories/Library/filtered-event-list/filtered-event-list";
 @import "./src/stories/Library/event-list-stacked/event-list-stacked";
+@import "./src/stories/Library/material-search/material-search";
 
 // Autosuggest block styling needs to be loaded before the rest of the scss for autosuggest
 @import "./src/stories/Blocks/autosuggest/autosuggest";
--- a/src/stories/Library/material-search/MaterialSearchLoading.tsx
+++ b/src/stories/Library/material-search/MaterialSearchLoading.tsx
@@ -1,0 +1,20 @@
+import LoadingLogo from "../../../public/icons/logo/reload_logo_black.svg";
+
+interface ReloadLoadingIconProps {
+  loadingText: string;
+}
+
+const MaterialSearchLoading = ({ loadingText }: ReloadLoadingIconProps) => {
+  return (
+    <div className="material-search__loading">
+      <img
+        src={LoadingLogo}
+        alt=""
+        className="material-search__loading-spinner"
+      />
+      <span className="material-search__loading-text">{loadingText}</span>
+    </div>
+  );
+};
+
+export default MaterialSearchLoading;
--- a/src/stories/Library/material-search/MaterialSearchPreview.tsx
+++ b/src/stories/Library/material-search/MaterialSearchPreview.tsx
@@ -1,0 +1,92 @@
+import { useState, useEffect } from "react";
+import Cover from "../cover/Cover";
+import { PreviewData } from "./MaterialSearchExampleData";
+import MaterialSearchLoading from "./MaterialSearchLoading"; // Assuming this is the correct import
+
+interface MaterialSearchPreviewProps {
+  displayMaterial: boolean;
+}
+
+const MaterialSearchPreview = ({
+  displayMaterial,
+}: MaterialSearchPreviewProps) => {
+  const [fictiveLoading, setFictiveLoading] = useState(false);
+
+  useEffect(() => {
+    if (displayMaterial) {
+      setFictiveLoading(true);
+      const timer = setTimeout(() => {
+        setFictiveLoading(false);
+      }, 1500);
+
+      return () => clearTimeout(timer);
+    }
+
+    return () => {};
+  }, [displayMaterial]);
+
+  if (fictiveLoading) {
+    return (
+      <div className="material-search__preview">
+        <div className="material-search__preview-loading">
+          <MaterialSearchLoading loadingText="Loading..." />
+        </div>
+      </div>
+    );
+  }
+
+  return (
+    <div className="material-search__preview">
+      {displayMaterial ? (
+        <div className="material-search__preview-material">
+          <Cover
+            src={PreviewData.coverUrl}
+            animate
+            size="large"
+            displaySize="small"
+          />
+          <div>
+            <div className="material-search__preview-item">
+              <span className="material-search__preview-term">Title:</span>
+              <span className="material-search__preview-detail">
+                {PreviewData.title}
+              </span>
+            </div>
+            <div className="material-search__preview-item">
+              <span className="material-search__preview-term">Author:</span>
+              <span className="material-search__preview-detail">
+                {PreviewData.author}
+              </span>
+            </div>
+            <div className="material-search__preview-item">
+              <span className="material-search__preview-term">
+                Publication Year:
+              </span>
+              <span className="material-search__preview-detail">
+                {PreviewData.publicationYear}
+              </span>
+            </div>
+            <div className="material-search__preview-item">
+              <span className="material-search__preview-term">Source:</span>
+              <span className="material-search__preview-detail">
+                {PreviewData.source}
+              </span>
+            </div>
+            <div className="material-search__preview-item">
+              <span className="material-search__preview-term">Work Id:</span>
+              <span className="material-search__preview-detail">
+                {PreviewData.workId}
+              </span>
+            </div>
+          </div>
+        </div>
+      ) : (
+        <div className="material-search__preview-empty">
+          No material selected
+        </div>
+      )}
+    </div>
+  );
+};
+
+export default MaterialSearchPreview;
--- a/src/stories/Library/material-search/material-search.scss
+++ b/src/stories/Library/material-search/material-search.scss
@@ -1,0 +1,223 @@
+// Variables
+$_gin_xs_spacing: 0.5rem;
+$_gin_m_spacing: 1rem;
+$_gin_xl_spacing: 2rem;
+$_gin_border_color: #8e929c;
+$_gin_font_weight_semibold: 525;
+$_gin-header-bg-color: #e1eafc;
+
+$_color__background-light: #eeeeed;
+$_color__hover-background: #8e929c;
+
+$_box-shadow: 0px 0px 0px 2px rgba(0, 0, 0, 0.25);
+$_loading-spinner-size: 40px;
+$_highlighted-item-bg-color: $_gin-header-bg-color;
+
+// Fixed preview height so the size doesn't jump if nothing is selected.
+$_preview-height: 170px;
+$_max-select-input-width: 220px;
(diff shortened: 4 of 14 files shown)


src/stories/Library/material-search/MaterialSearchPreview.tsx
```diff
@@ -0,0 +1,75 @@
...
+        <div className="material-search__preview-material">
+          <Cover src={PreviewData.coverUrl} animate size="large" />
+          <div className="material-search__preview-content">
+            <span className="material-search__preview-content__label">
+              Title:
+            </span>
+            <span>{PreviewData.title}</span>
+            <span className="material-search__preview-content__label">
+              Author:
+            </span>
```
kasperg, line 50: Consider reworking this to a [HTML description list](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl) instead of generic spans.

src/stories/Library/material-search/material-search.scss
```diff
@@ -0,0 +1,235 @@
+// Variables
+$_gin_xs_spacing: 0.5rem;
+$_gin_m_spacing: 1rem;
+$_gin_border_color: #8e929c;
+$_gin_font_weight_semibold: 525;
```
kasperg, line 5: :+1: Nice with these variables. The naming clarifies the relationship.

I wonder if it would be relevant to move these to a shared file where they can be used by more admin components.

src/stories/Library/material-search/material-search.scss
```diff
@@ -0,0 +1,235 @@
...
+// (I.e RecommendedMaterial) has its' prop set to "large", then image will
+// potentially not be displayed at all, since a "large" version is not served
+// by cover-service, but a small version is.
+
+// Arguably, this is a flaw in cover service, and should be fixed there, but
+// since there is no solution to this, and it potentially could happen for the
+// vast majority of materials I have decided to manually overwrite this here,
+// to ensure that the covers are displayed correctly and avoid frustration.
+$MATERIAL_SMALL_MOBILE: 104px;
+$MATERIAL_SMALL_DESKTOP: 137px;
```
kasperg, line 29: :+1: I appreciate this comment but I wonder how vast the problem actually is.

src/stories/Library/material-search/material-search.scss
```diff
@@ -0,0 +1,235 @@
...
+
+$list_header_height: 50px;
+
+// Material Search
+.material-search {
+  max-width: 800px;
```
kasperg, line 35: Consider documenting the origin of these fixed values.

There are some fixed values here - 800px, 220px, 170px - where I wonder what the relationship is to the rest if the layout.

src/stories/Library/material-search/material-search.scss
```diff
@@ -0,0 +1,235 @@
...
+  display: grid;
+  padding: 0 $_gin_m_spacing;
+  grid-template-columns: min-content auto;
+  column-gap: $_gin_m_spacing;
+  align-items: center;
+  height: 100%;
+
+  .cover {
+    height: $MATERIAL_SMALL_MOBILE;
+    @media screen and (min-width: 768px) {
```
kasperg, line 81: Please use the existing mixins for breakpoints.

If there is a reason they are not used then add a comment why.

src/stories/Library/material-search/material-search.scss
```diff
@@ -0,0 +1,235 @@
...
+  align-items: center;
+  justify-items: baseline;
+  text-align: left;
+  font-size: 14px;
+}
+
+.material-search__list-result-image-content img,
+.material-search__list-result-image-content .cover {
+  height: calc($MATERIAL_SMALL_MOBILE / 2);
+  @media screen and (min-width: 768px) {
```
kasperg, line 228: See comment about breakpoints elsewhere.

src/stories/Library/material-search/material-search.scss
```diff
@@ -0,0 +1,235 @@
...
+
+.material-search__preview-material {
+  display: grid;
+  padding: 0 $_gin_m_spacing;
+  grid-template-columns: min-content auto;
+  column-gap: $_gin_m_spacing;
+  align-items: center;
+  height: 100%;
+
+  .cover {
```
kasperg, line 79: Normally this nesting would not fly.

I recognize the need for it based on the comment in the top. I wonder if it should be fixed in the cover components through a modifier there instead of this nesting 🤔 .

src/stories/Library/material-search/MaterialSearchListResults.tsx
```diff
@@ -0,0 +1,62 @@
...
+  if (!showListResults) return null;
+
+  return (
+    <div className="material-search__list">
+      <div className="material-search__list-header">Total results: 5</div>
+      <ul className="material-search__list-results">
+        {ListResultData.map((work) => {
+          return (
+            <li key={work.workId}>
+              <button
```
kasperg, line 27: Using a `<button>` element is a good idea.

src/stories/Library/material-search/material-search.scss
```diff
@@ -0,0 +1,235 @@
...
+  flex-direction: column;
+  justify-content: left;
+  row-gap: $_gin_xs_spacing;
+  padding: $_gin_xs_spacing;
+  padding-bottom: 20px;
+  background-color: $_color__background-light;
+  margin: 0;
+}
+
+.material-search__list-result-button {
```
kasperg, line 179: Your BEM structure relies on putting everything in one block, `material-search` and then adding elements where the name implies more and more nesting, `list`, `list-results`, `list-result-button`. It is OK for you to break the element into more blocks.

kasperg (changes requested):
:+1: I enjoyed the small refactorings you made here. I have some comments for you to look into.

Our opinions differ when it comes to the design of the component. I would have expected the suggestions to appear like a dropdown directly below the input element but visually on top of any subsequent elements (e.g. the preview, other form elements etc.).

My preference is based on the argument that this would be similar to other form elements with suggestions where the relationship between the elements is clearer as they are adjacent instead of separated by the preview. It would also make the element take up less space visually and reduce layout shift as suggestions would appear on top of other elements.

Changing things are much too late at this point though.

src/stories/Library/material-search/MaterialSearchLoading.tsx
```diff
@@ -0,0 +1,22 @@
...
+const MaterialSearchLoading = ({ loadingText }: ReloadLoadingIconProps) => {
+  return (
+    <div className="material-search__loading">
+      <img
+        src={LoadingLogo}
+        alt="Loading..."
+        className="material-search__loading-spinner"
+      />
+      {loadingText && (
+        <span className="material-search__loading-text">{loadingText}</span>
```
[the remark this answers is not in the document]
LasseStaus: I've removed the alt entirely. 

The component itself is used 3 places, where one has loading text, and the others don't. 

I'm not sure what you mean by saying the text is always provided. I assume removing the alt text entirely is sufficient.

src/stories/Library/material-search/MaterialSearchPreview.tsx
LasseStaus: yeah thats better 👍

src/stories/Library/material-search/material-search.scss
LasseStaus: I have refactored cover to take a className being added to the cover wrapper. This way we can pass a specific BEM class to the component. I think this is simpler.

src/stories/Library/material-search/material-search.scss
LasseStaus: there isn't, oversight

src/stories/Library/material-search/material-search.scss
LasseStaus: Potentially yes, i think if we do maybe we could look into a bit more what is needed more broadly

src/stories/Library/material-search/MaterialSearchHiddenInputs.tsx
```diff
@@ -0,0 +1,42 @@
...
+const MaterialSearchHiddenInputs = ({
+  selectedWorkId,
+  selectedMaterialType,
+}: MaterialSearchHiddenInputsProps) => {
+  // These hidden inputs are used to store the selected work ID and material type in the drupal form
+  // In react, they are located in the storybook component, but in the drupal form they are created in the field widget and hidden with css.
+  return (
+    <div className="material-search__inputs-container">
+      <label className="material-search__label" htmlFor="hidden-work-id-input">
+        Hidden Work ID field
```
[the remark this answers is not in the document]
LasseStaus: I see what you mean. 
Since they are not a part of the actual app, i removed it from the material-search component. 

I think it adds value that they are there however, to see props/classes etc, but i moved it into the story instead, so they are still visible in storybook, but not a part of the markup for material-search, similar to what i did in the react repo. 

If you'd rather have it entirely removed let me know.

src/stories/Library/material-search/material-search.scss
LasseStaus: Not sure what to document really, these could be different since they are just a decision i made. If the designer wants it to change i think we should just change it then. I prefer the max width since it keeps the design compact, the alternative is that the components can grow to 1280px(gin styled container) which i think looks silly 🤷

base.scss
```diff
@@ -142,6 +142,7 @@
 @import "./src/stories/Library/opening-hours/opening-hours-skeleton";
 @import "./src/stories/Library/filtered-event-list/filtered-event-list";
 @import "./src/stories/Library/event-list-stacked/event-list-stacked";
+@import "./src/stories/Library/material-search/material-search";
```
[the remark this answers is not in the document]
LasseStaus: Yeah good point. 

We can look into this but i will move that out of scope of this ticket since it wasn't straightforward when i tried.

src/stories/Library/material-search/material-search.scss
kasperg: As I read the code now there is no need for mixins with breakpoints anymore.

kasperg (approved):

PR #641, opened 2024-05-27
Serve alternative images for card viewmodes. DDFFORM-667

rasben:
Card images are displayed in different aspect ratios, based on context. This is handled by object-fit centering, but generally speaking, we want different cropped images to be served.
We get around this, by allowing to pass along alternative SRCs, that are presented as containers, set to display none and using background images. By doing it this way, and letting CSS decide when the display is allowed, the browser is smart enough only to load the image when necessary.

https://reload.atlassian.net/browse/DDFFORM-667


src/stories/Library/card/CardImages.tsx
```diff
@@ -0,0 +1,47 @@
...
+                />
+              ))
+            : ""}
+
+          <img src={src} alt={alt} loading="lazy" />
+        </div>
+      ) : (
+        <div className="image-credited__no-image" />
+      )}
+    </figure>
```
LasseStaus, line 43: Maybe i am missing something, but i don't understand why the regular img is also rendered but hidden with css. 

If that is not necessary then you could avoid the empty string fallback and do: 

```
<div className="image-credited__image">
          {alternativeSrcs ? (
            alternativeSrcs.map((alternativeSrc) => (
              <div
                className="card__override-images"
                style={{
                  display: "none",
                  backgroundImage: `url(${alternativeSrc.src})`,
                }}
                data-card-style={alternativeSrc.name}
              />
            ))
          ) : (
            <img src={src} alt={alt} loading="lazy" />
          )}
  
```

If it is necessary, then avoid the string fallback with 

```
     {src ? (
        <div className="image-credited__image">
          {alternativeSrcs &&
            alternativeSrcs.map((alternativeSrc) => (
              <div
                className="card__override-images"
                style={{
                  display: "none",
                  backgroundImage: `url(${alternativeSrc.src})`,
                }}
                data-card-style={alternativeSrc.name}
              />
            ))}
          <img src={src} alt={alt} loading="lazy" />
        </div>
```

LasseStaus (approved):
Have tested on the deployment and it works nicely. 

See comment regarding whether the img needs to be rendered if there are alternativeSrcs before merging.

rasben:
It is necessary, so i'll update with your second example - thanks

PR #651, opened 2024-06-01
Add new `banner` component

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-540

#### Description

This pull request adds a new `banner` component that contains a required link and title. It can be used with a description and a background image to enhance visual appeal.

The `title` field is a `ReactNode` where the <u> (underline) tag has a custom underlined SVG placed beneath it.

#### Screenshot of the result

### DIFF
--- a/src/stories/Library/banner/Banner.stories.tsx
+++ b/src/stories/Library/banner/Banner.stories.tsx
@@ -47,12 +47,31 @@ export const Default = Template.bind({});
 
 export const NoImage = Template.bind({});
 NoImage.args = {
-  title: "Title <u>uden</u> billede",
   image: undefined,
 };
 
-export const NoDescription = Template.bind({});
-NoDescription.args = {
+export const NoImageOnlyTitle = Template.bind({});
+NoImageOnlyTitle.args = {
+  title: "Title <u>uden</u> billede",
+  image: undefined,
+  description: undefined,
+};
+
+export const NoImageOnlyDescription = Template.bind({});
+NoImageOnlyDescription.args = {
+  title: undefined,
+  image: undefined,
+  description:
+    "Om du er dedikeret musiknørd eller moderat musikinteresseret, så er dette siden til dig. Her kan du finde anbefalinger, digitale musikmagasiner, nyheder, musiklitteratur og meget mere.",
+};
+
+export const WithImageOnlyTitle = Template.bind({});
+WithImageOnlyTitle.args = {
   title: "Banner <u>uden</u> beskrivelse",
   description: undefined,
+};
+export const WithImageOnlyDescription = Template.bind({});
+WithImageOnlyDescription.args = {
+  description:
+    "Om du er dedikeret musiknørd eller moderat musikinteresseret, så er dette siden til dig. Her kan du finde anbefalinger, digitale musikmagasiner, nyheder, musiklitteratur og meget mere.",
 };
--- a/src/stories/Library/banner/Banner.stories.tsx
+++ b/src/stories/Library/banner/Banner.stories.tsx
@@ -1,0 +1,53 @@
+import { ComponentStory, ComponentMeta } from "@storybook/react";
+import { withDesign } from "storybook-addon-designs";
+import Banner from "./Banner";
+import ImageCredited from "../image-credited/ImageCredited";
+
+export default {
+  title: "Library / Banner",
+  component: Banner,
+  decorators: [withDesign],
+  parameters: {
+    design: {
+      type: "figma",
+      url: "https://www.figma.com/design/Zx9GrkFA3l4ISvyZD2q0Qi/Designsystem?node-id=446-6957&t=IWAOniAbcjV2y2Hf-4",
+    },
+  },
+  argTypes: {
+    image: {
+      defaultValue: (
+        <ImageCredited
+          src="https://images.unsplash.com/photo-1531058020387-3be344556be6?crop=entropy&cs=tinysrgb&fit=max&fm=jpg&ixid=MnwxfDB8MXxyYW5kb218MHx8ZXZlbnR8fHx8fHwxNzAyOTEwMzE0&ixlib=rb-4.0.3&q=8
+0&utm_campaign=api-credit&utm_medium=referral&utm_source=unsplash_source&w=1080"
+        />
+      ),
+    },
+    title: {
+      name: "Title",
+      defaultValue: "Hvad skal jeg <u>høre?</u>",
+      control: { type: "text" },
+    },
+    description: {
+      name: "Description",
+      defaultValue:
+        "Om du er dedikeret musiknørd eller moderat musikinteresseret, så er dette siden til dig. Her kan du finde anbefalinger, digitale musikmagasiner, nyheder, musiklitteratur og meget mere.",
+      control: { type: "text" },
+    },
+  },
+} as ComponentMeta<typeof Banner>;
+
+const Template: ComponentStory<typeof Banner> = (args) => <Banner {...args} />;
+
+export const Default = Template.bind({});
+
+export const NoImage = Template.bind({});
+NoImage.args = {
+  title: "Title <u>uden</u> billede",
+  image: undefined,
+};
+
+export const NoDescription = Template.bind({});
+NoDescription.args = {
+  title: "Banner <u>uden</u> beskrivelse",
+  description: undefined,
+};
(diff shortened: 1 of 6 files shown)


src/stories/Library/banner/banner.scss
```diff
@@ -0,0 +1,75 @@
...
+      position: absolute;
+      bottom: -5px;
+      left: 0;
+      width: 100%;
+      height: 10px;
+      background-size: 100% 100%;
+      // Underlined icon from public/icons/basic/icon-underlined.svg
+      background-image: url("data:image/svg+xml,%3Csvg width='101' height='7' viewBox='0 0 101 7' fill='none' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='M1.88867 6C36.7022 -0.286878 63.0591 0.988901 99.1109 1.35819' stroke='black' stroke-width='2' stroke-linecap='round' stroke-linejoin='round'/%3E%3C/svg%3E%0A");
+    }
+  }
```
kasperbirch1, line 56: Note: Look into making it more generic

src/stories/Library/banner/Banner.tsx
```diff
@@ -0,0 +1,40 @@
+import { FC, ReactNode } from "react";
+import clsx from "clsx";
+import MediaContainer from "../media-container/MediaContainer";
+import { ReactComponent as ArrowLargeRight } from "../Arrows/icon-arrow-ui/icon-arrow-ui-large-right.svg";
+
+type BannerType = {
+  title: string;
+  image?: ReactNode;
+  description?: string;
+};
```
LasseStaus, line 10: In the case that a link is required, i think this is missing a required link as prop.

src/stories/Library/banner/banner.scss
```diff
@@ -0,0 +1,65 @@
...
+  text-decoration: none;
+}
+
+.banner-content {
+  padding: $s-xl;
+  max-width: 375px;
+  background-color: white;
+  grid-row: 2;
+  grid-column: 1;
+
```
LasseStaus, line 29: In small mobile or less, there is no space on right side. I think you could do something likes, up to you wheter to use % or something else. Remember to change it on larger sizes if necessary. 

```suggestion
.banner-content {
  padding: $s-xl;
  max-width: 375px;
  background-color: white;
  grid-row: 2;
  grid-column: 1;
  width: calc(100% - 10%);
```

src/stories/Library/banner/banner.scss
```diff
@@ -0,0 +1,65 @@
...
+.banner-visual {
+  grid-row: 1 / -1;
+  grid-column: 1;
+
+  img {
+    @include banner-height;
+    object-fit: cover;
+    object-position: center;
+  }
+}
```
LasseStaus, line 65: I'm unsure why this is needed. 
there was a todo " // Todo: can this be done in drupal?" which i am not sure what ment

The way i see it, if you choose to use the media-container approach, these styles are already applied besides the height, which can be put on another element if necessary. 

```suggestion
.banner-visual {
  grid-row: 1 / -1;
  grid-column: 1;
}
```

src/stories/Library/banner/banner.scss
```diff
@@ -0,0 +1,65 @@
...
+    height: 810px;
+  }
+}
+
+@mixin banner-text-margin {
+  margin-bottom: $s-md;
+  @include media-query__medium {
+    margin-bottom: $s-xl;
+  }
+}
```
LasseStaus, line 13: Not sure if a mixin for this is necessary. 

suggestion for naming: banner-text-spacing 

Alternatively just use gap with grid/flex? 

up to you though.

src/stories/Library/banner/banner.scss
```diff
@@ -0,0 +1,65 @@
...
+  padding: $s-xl;
+  max-width: 375px;
+  background-color: white;
+  grid-row: 2;
+  grid-column: 1;
+
+  @include media-query__medium {
+    padding: $s-2xl;
+    max-width: 575px;
+    margin-left: 250px;
```
LasseStaus, line 33: I personally don't think this is a good approach to positioning the text container. The value of 250px might be correct in the figma example, but as the screen changes i would argue this should change too, and in that case this isn't responsive

I would suggest you to place the text inside a container with a smaller layout-container, and align that left side to make it move responsively.

LasseStaus (changes requested):
Good stuff, i have some suggestions. 

I think it could be considered if a link is required. Maybe ddf wants the component to also be used without a link. This is potentially out-of-scope here since they didn't specify it, but i think its' worth asking if we haven't. 


I'm unsure about the implementation of the image. By doing this, we use everything related to media-container, which has figures, fallbacks etc which might not be necessary. 

I would suggest you consider setting the image as a backgrundImage directly inline instead, and simplify the html structure. This is also being done [here](https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/641). Maybe @rasben could chime in what he thinks about that. 


I'd suggest you link to related PR's also for convenience 🚀 

Good stuff with the underline!

src/stories/Library/banner/banner.scss
kasperbirch1: This is necessary because we render images in a 16:9 aspect ratio in Drupal.

src/stories/Library/banner/banner.scss
kasperbirch1: I personally liked it, but I'm also open to using the grid if you think it's better.

src/stories/Library/banner/banner.scss
kasperbirch1: Thanks 🙏

src/stories/Library/banner/banner.scss
kasperbirch1: Changed my opinion as it will place the arrow incorrectly if there is no describtion

kasperg (approved):
:+1: I think this is nice. I have some questions to be considered but it may be too late to do much about them at this point.

There seems to be some edge cases where the component does not appear as well. I do not think that is a major problem though.

Banner - With Image Only Description ⋅ Storybook 2024-06-20 09-00-24

src/styles/scss/tools/mixins.tools.scss
```diff
@@ -70,3 +70,27 @@
...
+    &::after {
+      content: "";
+      position: absolute;
+      bottom: -5px;
+      left: 0;
+      width: 100%;
+      height: 10px;
+      background-size: 100% 100%;
+      // Underlined icon from public/icons/basic/icon-underlined.svg
+      background-image: url("data:image/svg+xml,%3Csvg width='205' height='8' viewBox='0 0 205 8' fill='none' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='M1.11133 6.87425C73.7223 0.129386 128.695 1.4981 203.889 1.8943' stroke='black' stroke-width='2' stroke-linecap='round' stroke-linejoin='round'/%3E%3C/svg%3E%0A");
```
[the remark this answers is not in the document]
LasseStaus: Yea we should look into doing that. Currently this is being done a few places. I would also say it is out of scope for this 👍

src/stories/Library/banner/Banner.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+  },
+  argTypes: {
+    image: {
+      defaultValue: (
+        <ImageCredited
+          src="https://images.unsplash.com/photo-1531058020387-3be344556be6?crop=entropy&cs=tinysrgb&fit=max&fm=jpg&ixid=MnwxfDB8MXxyYW5kb218MHx8ZXZlbnR8fHx8fHwxNzAyOTEwMzE0&ixlib=rb-4.0.3&q=8
+0&utm_campaign=api-credit&utm_medium=referral&utm_source=unsplash_source&w=1080"
+        />
+      ),
+    },
```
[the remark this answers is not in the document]
LasseStaus: This is changed as a result of rewriting to background image

src/stories/Library/banner/Banner.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+const Banner: FC<BannerType> = ({ link, image, title, description }) => {
+  return (
+    <a href={link} className="banner">
+      {image && (
+        <div className="banner__media-wrapper">
+          <MediaContainer media={image} />
+        </div>
+      )}
```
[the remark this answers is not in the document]
LasseStaus: I agree. 

Have rewritten to using backgroundImage inline

src/stories/Library/banner/Banner.tsx
```diff
@@ -0,0 +1,47 @@
...
+      <div
+        className={clsx("banner__content-wrapper", {
+          "banner__content-wrapper--no-image": !image,
+        })}
+      >
+        <div
+          className={clsx("banner__content arrow__hover--right-large", {
+            "banner__content--no-image": !image,
+          })}
+        >
```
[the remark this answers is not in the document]
LasseStaus: True i missed that actually. I think it should be on the main element and not this `banner__content`. Have moved it.

LasseStaus (approved):

PR #657, opened 2024-06-04
Review inputs to form-styling.

rasben:
´

### DIFF
--- a/docs/architecture/adr-002-form-styling.md
+++ b/docs/architecture/adr-002-form-styling.md
@@ -34,6 +34,8 @@ expanding the current `dpl-form` class.
 This currently has an exception in form of form inputs - these have been styled
 a long time ago and use the class `dpl-input`.
 
-### Implementation in the dpl-design-system
+@todo 👆can we do something about this `dpl-input`?
 
-Here is the [link to our form css file](../src/stories/Blocks/form/form.scss).
+### Implementation
+
+See [form.scss for our implementation](../../src/stories/Blocks/form/form.scss).
--- a/src/stories/Blocks/form/form.scss
+++ b/src/stories/Blocks/form/form.scss
@@ -5,55 +5,57 @@
   padding-left: $s-md;
   padding-right: $s-md;
   box-sizing: border-box;
+}
 
-  &__header {
-    margin: $s-2xl 0 $s-2xl $s-md;
+.dpl-form__header {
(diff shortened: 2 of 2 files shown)


src/stories/Blocks/form/form.scss
```diff
@@ -5,55 +5,57 @@
   padding-left: $s-md;
   padding-right: $s-md;
   box-sizing: border-box;
+}
 
-  &__header {
-    margin: $s-2xl 0 $s-2xl $s-md;
+.dpl-form__header {
```
Adamik10, line 10: Mmmmm, 🤔 why?

docs/architecture/adr-002-form-styling.md
```diff
@@ -34,6 +34,8 @@ expanding the current `dpl-form` class.
 This currently has an exception in form of form inputs - these have been styled
 a long time ago and use the class `dpl-input`.
 
-### Implementation in the dpl-design-system
+@todo 👆can we do something about this `dpl-input`?
 
-Here is the [link to our form css file](../src/stories/Blocks/form/form.scss).
+### Implementation
+
+See [form.scss for our implementation](../../src/stories/Blocks/form/form.scss).
```
Adamik10, line 41: Danke 🙏🏼

Adamik10 (changes requested):
Generally, I don't see the point in changing nested classes into separate definitions.
+ something must have happened, because there's a bunch of Chromatic changes.
But I really enjoy you adding that TODO + fixing up the link.

Should I just take what I like from this PR and make my own, or do you wanna take this one further?

src/stories/Blocks/form/form.scss
rasben: https://github.com/danskernesdigitalebibliotek/dpl-design-system/blob/develop/docs/scss.md#bem-naming-convention

>(&__ should be avoided, to avoid massive indention.)

This is something we have started to do, moving forward.
Another value, apart from avoiding indention, is also that when you use `&__` you make it really hard to debug things.

Let's say you have a class called `dpl-form__header` that you can see in the browser, and you want to find it in the site files - if you just search for "dpl-form__header" it wont show up.

rasben:
>**Should I just take what I like from this PR and make my own**, or do you wanna take this one further?

Yes, please, make your own PR :)

PR #658, opened 2024-06-06
Update the Multiselect component [DDFLSBP-601]

clausbruun:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-601

#### Description

Update the Multiselect component and avoid the default blue button color on iOS.


Adamik10 (approved):
Lovely :)

Do these changes also need to be implemented in the react repository? Or is it already a button there?

PR #662, opened 2024-06-13
Fix SCSS variables [DDFLSBP-660]

clausbruun:
#### Link to issue

https://reload.atlassian.net/browse/DDFLSBP-660

#### Description

Correct SCSS variable interpolation for CSS variable and add font-body CSS variable for use in webforms.

#### Sibling PR

https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1255


Dresse (approved):
LGTM 👍 

I'm guessing this change is what makes it possible to use the variables in the CMS?

PR #668, opened 2024-06-17
Feature/find on shelf changes

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-717

#### Description
Design was breaking **on small desktops**. These changes fix that.

#### Screenshot of the result
BEFORE:
image

AFTER:
image

Sibling PR


Dresse (approved):
LGTM - However I have a question. The availability labels is still cutoff. Should we consider fixing that here as well? Or should this be left alone/be a separate task?

PR #707, opened 2024-08-14
Add drupal checkbox and radio element styling stories

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFLSBP-647

#### Description
This PR introduces Drupal checkbox specific stories + classes that override the Drupal checkbox styling to match the project design system.

#### Screenshot of the result
image
image

Sibling PR in Drupal


src/stories/Library/Forms/checkbox/checkbox.scss
```diff
@@ -122,4 +124,8 @@
     margin: 0;
     height: auto;
   }
+
+  .dpl-input.checkbox-override .input-label {
+    cursor: pointer;
```
Adamik10, line 129: @Dresse would hitting the label actually trigger the checkbox to get checked and unchecked? I intentionally left it without cursor pointer because I thought that in Drupal we can't get that functionality. So I just thought it would be confusing for the user to think they can click the label but then nothing actually happens.

Dresse: Yes, clicking the label is checking/unchecking the checkbox. The same goes for the radio buttons 😄

src/stories/Library/Forms/radio/Radio.stories.tsx
```diff
@@ -0,0 +1,36 @@
...
+} as ComponentMeta<typeof RadioDrupal>;
+
+const Template: ComponentStory<typeof RadioDrupal> = (args) => (
+  <RadioDrupal {...args} />
+);
+
+export const drupalRadio = Template.bind({});
+drupalRadio.args = {
+  legend: "Try this radio",
+  description: "| This is a description.",
```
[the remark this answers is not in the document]
Dresse: I agree Thomas. I have removed it for now but haven't added it as part of the UI as I think it looks okay even without it.
We can revisit it if DDF wants to have some kind of divider.

Dresse (approved):
LGTM

PR #708, opened 2024-08-14
Unify lists between cms and react

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFFORM-953

#### Description
This pull request builds on top of [PR #699](https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/699) in the dpl-design-system repository. The primary goal is to streamline the CSS styles for lists between the Drupal and React applications.

##### Depend on: 
https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/699

#### Test 
~~https://varnish.pr-1484.dpl-cms.dplplat01.dpl.reload.dk/search?q=dans~~
https://varnish.pr-1497.dpl-cms.dplplat01.dpl.reload.dk/search?q=dans


src/stories/Blocks/content-list-page/content-list-page.scss
```diff
@@ -19,6 +19,7 @@
   }
 }
 
+//  Todo should be merged with .facet-line
```
xendk, line 22: Just a comment: I don't buy into the "no todos in code!" mindset, but: a todo should either explain what and why something needs to be done or link to an issue that explains it. Otherwise, it becomes a personal todo that only you know the reason for. And my personal experience is that, two weeks from now, I can't remember why I thought they should be merged.

If the same todo applies in multiple places, they can point to each other to avoid redundancy.

xendk (approved):
Looks good to me. :+1:

PR #736, opened 2024-09-30
Add underline styling for link-tags. DDFFHER-41

rasben:
/?path=/story/library-tag-list--with-tags

Screenshot 2024-09-30 at 15 25 54


kasperbirch1 (approved):
LGTM 👍

PR #741, opened 2024-10-07
Fix "pause reservations" popup

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFBRA-52

#### Description
This PR fixes an issue where the date picker for pausing reservations was bleeding outside the screen on small devices. It is possible to scroll, but the users couldn't figure it out because one needs to scroll inside the date picker overlay window. But now they should be able to see the whole window at all times.

#### Screenshot of the result
AFTER:
image

BEFORE:
image

-

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
     "react-dom": "^17.0.2",
     "react-helmet": "^6.1.0",
     "react-scripts": "^4.0.3",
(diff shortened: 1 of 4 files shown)


Dresse (approved):
LGTM 👍

PR #744, opened 2024-10-15
DDFHER-96 - Add `@media print` to enable printing of articles in modals

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFHER-96

#### Description
This pull request adds `@media print` CSS styles to our modal component to enable proper printing of articles displayed within modals. While this solution addresses the immediate issue, we might want to consider using a library like `react-to-print` for a more beautiful approach in the future.

#### Screenshot of the result
[print.pdf](https://github.com/user-attachments/files/17375746/print.pdf)

#### Test
https://varnish.pr-1657.dpl-cms.dplplat01.dpl.reload.dk/work/work-of%3A870971-avis%3A89724244?type=artikel&modal=infomedia-modal-870971-avis%3A89724244

### DIFF
--- a/src/stories/Library/Modals/modal.scss
+++ b/src/stories/Library/Modals/modal.scss
@@ -119,18 +119,18 @@
 // Adjust the modal to display all content when printing
 @media print {
   .modal {
-    position: static !important;
-    overflow: visible !important;
-    width: auto !important;
-    height: auto !important;
-    opacity: 1 !important;
-    display: block !important;
+    position: static;
+    overflow: visible;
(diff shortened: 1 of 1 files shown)


kasperg (approved):
:+1:

I am not too happy about the result of this change. We went from a print version which only displays the content of the modal but also only displays the first page multiple time to a print version which displays the entire page including both the modal and the underlying content.

We have both tried to address this in multiple ways but without succeeding at a better result. We have decided to proceed with the suggested change for now.

PR #749, opened 2024-10-23
Upgrade node to v20

JacobArrow:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-137

#### Description

This PR upgrades node to version 20.

Be aware that we temporarily need to run storybook with some extra node options as a workaround in v6.

Run `nvm install --lts` and `nvm use` to get started with node 20.


kasperg (approved):
:+1: 🎉

PR #750, opened 2024-10-24
DDFHER-98 - Hide the drupal "Apply" button in `search-full-text` form

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFHER-98
https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/1678

#### Description
The "Apply" button in the Drupal form cannot be removed directly in the view, so it is hidden using CSS to prevent it from displaying.

**Before:**
Skærmbillede 2024-10-24 kl  11 00 44
**After:**
Skærmbillede 2024-10-24 kl  11 01 45

### DIFF
--- a/src/stories/Library/search-full-text/search-full-text.scss
+++ b/src/stories/Library/search-full-text/search-full-text.scss
@@ -26,3 +26,9 @@
   display: grid;
   place-items: center;
 }
+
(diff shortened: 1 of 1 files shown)


kasperbirch1:
We moved the CSS inside the base.css we have in Drupal CMS.

PR #796, opened 2024-12-05
KBHBIB-24 Added new share buttons

JacobArrow:
#### Link to issue

Please add a link to the issue being addressed by this change.

#### Description

- Added share button component with inline and fixed buttons for sharing current page on Facebook or copying the URL to clipboard

#### Screenshot of the result

Fixed buttons
image

Inline buttons
image


src/stories/Library/Buttons/button-share/button-share.tsx
```diff
@@ -0,0 +1,53 @@
...
+        <button
+          type="button"
+          onClick={() => navigator.clipboard.writeText(href)}
+          aria-label="Kopier denne side til udklipsholderen"
+          className="button-share__button button-share__button--fixed"
+        >
+          <LinkSvg role="img" aria-hidden="true" />
+        </button>
+      </div>
+      <a
```
Adamik10, line 34: We have a `<LinkButton />` component and a `<Links />` component that we normally use for links like this :) Please use one of those depending on which one fits better.

src/stories/Library/Buttons/button-share/button-share.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+export const ButtonShare = ({
+  href,
+  textFacebook,
+  textCopy,
+}: ButtonShareProps) => {
+  return (
+    <div className="button-share">
+      <div className="button-share button-share--fixed">
+        <a
```
Adamik10, line 18: We have a `<LinkButton />` component in the repository. Utilize that here, please. (if it doesn't fit for some reason, we also have a `<Links />` component for links (<a/> tags) (not sure why it's defined in plural 🤷)

Adamik10 (commented):
I have a couple of comments that I'd like you to look at. 
Also, is there a third PR for the dpl-react repo? Or have I already looked at that one and just can't remember?

src/stories/Library/Buttons/button-share/button-share.tsx
JacobArrow: `<Links />` takes a string as link text, but I need to render an SVG, that it why I cannot use the `<Links />` component

src/stories/Library/Buttons/button-share/button-share.tsx
JacobArrow: `<Links />` takes a string as link text, but I need to render an SVG, that it why I cannot use the `<Links />` component

PR #809, opened 2024-12-18
DDFBRA-225 - Use `svh` height to ensure consistent modal layout

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFBRA-225

#### Description

Ensures a more consistent layout on mobile devices by accounting for dynamic changes in viewport height caused by browser UI elements, such as the address bar in Chrome.

#### Test
https://varnish.pr-1877.dpl-cms.dplplat01.dpl.reload.dk/

https://github.com/user-attachments/assets/f07bf82e-b125-453c-b36f-f4e8e60b705e

### DIFF
--- a/src/stories/Library/Modals/modal.scss
+++ b/src/stories/Library/Modals/modal.scss
@@ -8,7 +8,7 @@
   opacity: 0;
   display: none;
   transition: opacity 0.3s;
-  height: 100vh;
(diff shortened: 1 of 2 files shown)


Adamik10 (approved):
We clap and we cheer

PR #828, opened 2025-01-20
Center calendar horizontally on mobile and hide box arrow

JacobArrow:
#### Link to issue

https://reload.atlassian.net/browse/DDFBRA-364

#### Description

It seems the previous CSS rule was targeting a too small device size, as many phones are larger than 375px. 
I also tweaked the CSS, so the calendar is centered horizontally. 
As for the issue with scrolling, it should have been fixed in #809

#### Screenshot of the result
image


Adamik10 (approved):
✨

PR #890, opened 2025-04-30
DDFSAL 1 DDFSAL 4

LasseStaus:
- **Update MaterialGrid structure**
- **Add MaterialGridRelated with CSS & Story**

#### Link to issue

[DDFSAL-1](https://reload.atlassian.net/browse/DDFSAL-1)
[DDFSAL-4](https://reload.atlassian.net/browse/DDFSAL-4)

[design system](https://github.com/danskernesdigitalebibliotek/dpl-design-system/pull/890)
[react repo](https://github.com/danskernesdigitalebibliotek/dpl-react/pull/1821)
[CMS](https://github.com/danskernesdigitalebibliotek/dpl-cms/pull/2344)

#### Description

Please include a short description of the suggested change and the reasoning behind the approach you have chosen.

Stories are left unaccepted until pr is reviewed. 

#### Screenshot of the result

image

[DDFSAL-1]: https://reload.atlassian.net/browse/DDFSAL-1?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DDFSAL-4]: https://reload.atlassian.net/browse/DDFSAL-4?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/stories/Library/material-grid-related/material-grid-related.scss
+++ b/src/stories/Library/material-grid-related/material-grid-related.scss
@@ -1,0 +1,88 @@
+.material-grid-related {
+  margin-bottom: 1rem;
+}
+
+.material-grid-related__header {
+  display: flex;
+  flex-direction: column;
+  gap: 1rem;
+  margin-bottom: 20px;
+  @include layout-container($layout__max-width--large);
+
+  @include media-query__small {
+    flex-direction: row;
+    justify-content: space-between;
+    align-items: center;
+    margin-bottom: 30px;
+  }
+}
+
+.material-grid-related__title {
+  @include typography($typo__h2);
+}
+
+.material-grid-related__dropdown {
+  display: inline-flex;
+  align-items: center;
+
+  @include media-query__small {
+    display: none;
+  }
+}
+
+.material-grid-related__filter-inline {
+  display: none;
+  margin-top: auto;
+  gap: $s-md;
+
+  @include media-query__small {
+    display: flex;
+  }
+}
+
+.material-grid-related__filter-button {
+  background: none;
+  border: none;
+  cursor: pointer;
+  padding: 0;
+  font-weight: $font__weight--normal;
+  color: inherit;
+  position: relative;
+  display: inline-flex;
+  align-items: flex-start;
+  gap: 8px;
+}
+
+.material-grid-related__filter-text {
+  position: relative;
+  display: inline-block;
+  text-transform: uppercase;
+  @include typography($typo__body-small);
+
+  &::after {
+    content: "";
+    position: absolute;
+    left: 0;
+    bottom: -5px;
+    width: 100%;
+    height: 1px;
+    background-color: $color__global-black;
+  }
+
+  &:hover {
+    text-shadow: 0 0 0.2px currentColor;
+  }
+}
+.material-grid-related__filter-sup {
+  position: relative;
+  top: -50%;
+}
+
+.material-grid-related__filter-button--active {
+  // Text-shadow is used for bold effect that won't move the text
+  text-shadow: 0 0 0.4px currentColor;
+
+  .material-grid-related__filter-text::after {
+    height: 2px;
+  }
+}
--- a/src/stories/Library/material-grid-related/material-grid-related.tsx
+++ b/src/stories/Library/material-grid-related/material-grid-related.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { ReactComponent as TriangleIcon } from "../../../public/icons/basic/icon-triangle.svg";
+import { Dropdown } from "../dropdown/Dropdown";
 import { MaterialGrid } from "../material-grid/MaterialGrid";
 import MaterialGridData from "../material-grid/MaterialGridData";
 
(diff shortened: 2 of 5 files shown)


src/stories/Library/material-grid-related/material-grid-related.tsx
```diff
@@ -0,0 +1,73 @@
...
+            <option value="series">Series</option>
+            <option value="author">Author</option>
+          </select>
+          <div className="dropdown__arrows dropdown__arrows--inline">
+            <TriangleIcon
+              className="dropdown__arrow dropdown__arrow--bottom"
+              aria-hidden="true"
+            />
+          </div>
+        </div>
```
kasperg, line 28: Please consider if this could be reduced to use the existing `Dropdown` component.

src/stories/Library/material-grid-related/material-grid-related.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { ReactComponent as TriangleIcon } from "../../../public/icons/basic/icon-triangle.svg";
+import { MaterialGrid } from "../material-grid/MaterialGrid";
+import MaterialGridData from "../material-grid/MaterialGridData";
+
+const MaterialGridRelated: React.FC = () => {
+  return (
+    <div className="material-grid-related">
+      <div className="material-grid-related__header">
+        <h2 className="material-grid-related__title">Andre materialer</h2>
+        {/* Mobile display : Dropdown */}
```
kasperg, line 11: :+1: This is helpful. 

As I read the code we are using break points to set `display: none` on the alternate version which should prevent screen readers and the like from rendering the component twice.

If you are uncertain about whether this is the case then it is probably worth double checking :smile:.

src/stories/Library/material-grid-related/material-grid-related.tsx
```diff
@@ -0,0 +1,73 @@
...
+              className="dropdown__arrow dropdown__arrow--bottom"
+              aria-hidden="true"
+            />
+          </div>
+        </div>
+        {/* Desktop display : inline */}
+        <div
+          className="material-grid-related__filter-inline"
+          role="group"
+          aria-label="Filter displayed materials"
```
kasperg, line 33: :+1: for adding an ARIA label here. I think it makes sense structurally even though it is not shown visually.

I did not know about the `group` role. Frankly I do not understand [the documentation at MDN either](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/group_role). I hope it is something you put an effort into.

src/stories/Library/material-grid-related/material-grid-related.tsx
```diff
@@ -0,0 +1,73 @@
...
+        >
+          <button
+            className="material-grid-related__filter-button material-grid-related__filter-button--active"
+            aria-pressed="true"
+          >
+            <span className="material-grid-related__filter-text">
+              Recommendations
+            </span>
+            <sup>8</sup>
```
kasperg, line 42: Please consider if using `<sup>` without further styling here achieves the intended design.

When comparing Figma with Storybook the position of the numbers in Storybook do not seem right to me.

src/stories/Library/material-grid-related/material-grid-related.tsx
```diff
@@ -0,0 +1,73 @@
...
+            />
+          </div>
+        </div>
+        {/* Desktop display : inline */}
+        <div
+          className="material-grid-related__filter-inline"
+          role="group"
+          aria-label="Filter displayed materials"
+        >
+          <button
```
kasperg, line 35: :+1: for using buttons here.

src/stories/Library/material-grid-related/material-grid-related.scss
```diff
@@ -0,0 +1,78 @@
...
+  color: inherit;
+  position: relative;
+  display: inline-flex;
+  align-items: flex-start;
+  gap: 8px;
+}
+
+.material-grid-related__filter-text {
+  position: relative;
+  display: inline-block;
```
kasperg, line 58: Please consider adding `text-transform: uppercase;` here.

As I see it the casing of the filers are part of the design, not a translation.

src/stories/Library/material-grid-related/material-grid-related.tsx
```diff
@@ -0,0 +1,73 @@
...
+          <div className="dropdown__arrows dropdown__arrows--inline">
+            <TriangleIcon
+              className="dropdown__arrow dropdown__arrow--bottom"
+              aria-hidden="true"
+            />
+          </div>
+        </div>
+        {/* Desktop display : inline */}
+        <div
+          className="material-grid-related__filter-inline"
```
kasperg, line 31: Consider if this is related to the [Link filters component](https://616ffdab9acbf5003ad5fd2b-vgmirlovej.chromatic.com/?path=/story/library-links-link-filters--item).

To me it seems to be the case. I do not know if it useful in any way though.

kasperg (changes requested):
:+1: Nice changes here.

I have a few comments and a couple of changes that I think should be addressed.

src/stories/Library/material-grid-related/material-grid-related.tsx
LasseStaus: I think it is reasonable to see this as a the Aria version of a legend, Fieldset or similar. 

A group of controls that are realted but not a major section of a page. 👍

src/stories/Library/material-grid-related/material-grid-related.tsx
LasseStaus: As i see it the link filters component deals with nav-links, and not button presses. 

I also do not see a "link filters" component equivalent of Dropdown or similar, and looking at the link filters usage i don't think i should mix buttons & aria into it right now for this case. 🤷

src/stories/Library/material-grid-related/material-grid-related.tsx
LasseStaus: I've updated the styles a bit to match better. 👍. 

Think i missed it, after missing that the text should be standard uppercase. Gaps were fine tho

PR #907, opened 2025-05-29
Add autosuggest-backdrop styling

Adamik10:
#### Link to issue
https://reload.atlassian.net/browse/DDFBRA-646

#### Description
This PR adds styling for autosuggest backdrop

#### Screenshot of the result
-

#### Additional comments or questions
-

PR #967, opened 2025-08-11
DDFSAL-296 - Correct header sidebar navigation z-index

kasperbirch1:
#### Link to issue
https://reload.atlassian.net/browse/DDFSAL-296

#### Description

Ensures the sidebar navigation uses the right z-index. Addresses a regression from PR #907, aligning with the fix made for the facet browser in PR #961.


ThomasGross (approved):
Looks good 👍

